PR #2, opened 2019-02-26
Refactor of Java CGOL example

falkecarlsen:
Refactored object encapsulation of cell away and local optimisations result in 36% load reduction

### DIFF
--- a/game_of_life-java/src/main/java/GameOfLife/Canvas.java
+++ b/game_of_life-java/src/main/java/GameOfLife/Canvas.java
@@ -37,7 +37,7 @@ class Canvas {
                 neighbours = countMooreNeighbours(i, j);
 
                 // set new state in tempCanvas based on the life of the given cell in canvas
-                tempCanvas[i][j]=(cellLife(canvas[i][j], neighbours));
+                tempCanvas[i][j] = (cellLife(canvas[i][j], neighbours));
             }
         }
         // swap references to arrays
--- a/game_of_life-java/src/main/java/GameOfLife/Canvas.java
+++ b/game_of_life-java/src/main/java/GameOfLife/Canvas.java
@@ -6,24 +6,24 @@ import java.util.Random;
 class Canvas {
     private final int x;
     private final int y;
-    private Cell[][] canvas;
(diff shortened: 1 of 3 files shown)


game_of_life-java/src/main/java/GameOfLife/Canvas.java
```diff
@@ -6,38 +6,38 @@
...
         // temp canvas-array for swapping before return
-        Cell[][] temp;
+        boolean[][] temp;
         for (int i = 0; i < this.x; i++) {
             for (int j = 0; j < this.y; j++) {
                 neighbours = countMooreNeighbours(i, j);
 
                 // set new state in tempCanvas based on the life of the given cell in canvas
-                tempCanvas[i][j].setState(cellLife(canvas[i][j], neighbours));
+                tempCanvas[i][j]=(cellLife(canvas[i][j], neighbours));
```
jenrik: missing space around assignment operator

game_of_life-java/src/main/java/GameOfLife/Graphics.java
```diff
@@ -24,12 +24,14 @@ public void windowClosing(WindowEvent windowEvent) {
     }
 
     public static void drawCanvas(Canvas canvas, Frame frame) {
+        boolean[][] grid = canvas.getCanvas();
+        java.awt.Graphics graphics = frame.getGraphics();
```
jenrik, line 28: Why isn't `java.awt.Graphics` imported?

jenrik (changes requested):

game_of_life-java/src/main/java/GameOfLife/Canvas.java
falkecarlsen: Fixed in d984780

game_of_life-java/src/main/java/GameOfLife/Graphics.java
falkecarlsen: It certainly is in line 3, however, the class is called `Graphics, the FQN of the class must be used to differentiate.

PR #5, opened 2019-03-04
Grammar feature expansion and tidying

falkecarlsen:
Refactored `switch` and `func` declarations from grammar. Also renamed `board` to `world`, both as syntax and internally in grammar definition.

Also added support for custom neighbourhood definitions, RGB-notation of colour for each state, and two built-in functions; `count(neighbourhood, state)` and `rand(max)`.

- `example.cell` has been revised to conform to the new grammar, by removing return-statement
- `example2.cell` has been revised to conform to the new grammar, by removing function-, return-, and switch-usage
- `cgol.cell` Note that custom neighbourhoods are yet to be defined by the grammar.

### DIFF
--- a/cellumata_syntax-v2/cellmata.g4
+++ b/cellumata_syntax-v2/cellmata.g4
@@ -1,16 +1,17 @@
 grammar cellmata;
 
-start : board_decl body EOF;
+start : world_dcl body EOF;
 
 body : (state_decl | const_decl)*;
 const_decl : 'const' const_ident ASSIGN expr ;
 const_ident : IDENT ;
 
 // Board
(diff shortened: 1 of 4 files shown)


jenrik (approved):

NicEastvillage (changes requested):
Why is cellsize deleted in the examples?

falkecarlsen:
@NicEastvillage cellsize is now allowed in grammar and has been added back to `cgol.cell`.

NicEastvillage (approved):
One thing: Check if all the examples are correct regarding the "new" order of definitions in world declaration. Otherwise, aproved.

PR #6, opened 2019-03-07
Remove support for infinite dimension in grammar and refactored example*.cell into one .cell

falkecarlsen:
- Removed support for `infinite` definition of a dimension bound.
- Refactored `example.cell` and `example2.cell` into `syntactically-correct.cell` which aims to employ as many syntactically correct productions as possible. This has been moved to the directory `syntax_testing` anticipating more examples for testing grammar.


cellumata_syntax-v2/cellmata.g4
```diff
@@ -23,7 +23,7 @@ code_block : BLOCK_START stmt* BLOCK_END ;
 stmt : (if_stmt | become_stmt | assign_stmt | increment_stmt | decrement_stmt) ;
 
 assign_stmt : 'let'? (var_ident | array_lookup) ASSIGN expr END ;
-if_stmt : STMT_IF PAREN_START expr PAREN_END code_block (STMT_ELSE STMT_IF PAREN_START expr PAREN_END code_block)* (STMT_ELSE code_block)? ;
+if_stmt : STMT_IF PAREN_START expr PAREN_END (stmt | code_block) (STMT_ELSE STMT_IF PAREN_START expr PAREN_END (stmt | code_block))* (STMT_ELSE (stmt | code_block))? ;
```
jenrik: Allowing if statements without braces increases the likely hood a programmer would incorrectly assume a second statement is also gated by the condition, I there disagree with the change

jenrik (changes requested):

cellumata_syntax-v2/cellmata.g4
falkecarlsen: Agreed. However, I recall that we agreed on supporting this syntax during group-discussions. Let's continue that talk next week.

falkecarlsen: Relevant commit has been reverted.

jenrik (approved):

PR #12, opened 2019-03-13
Updated expression syntax in accordance with the abstract syntax

jenrik:
In accordance with the abstract syntax the following have been change:

- Removed xor operator
- Changed the inequality operator from `is not` to `!=`
- Changed the equality operator from `is` to `==`
- Changed the inversion operator from `not` to `!`
- Changed the and operator from `and` to `&&`
- Changed the or operator from `or` to `||`

### DIFF
--- a/parser/cellmata.g4
+++ b/parser/cellmata.g4
@@ -162,7 +162,7 @@ PAREN_END : ')' ;
 END : ';' ;
 
 OP_COMPARE : '==' ;
-OP_COMPARE_not : '!=' ;
+OP_COMPARE_NOT : '!=' ;
 OP_NOT : '!' ;
 OP_INCREMENT : '++' ;
 OP_DECREMENT : '--' ;
--- a/parser/cellmata.g4
+++ b/parser/cellmata.g4
@@ -83,11 +83,7 @@ bool_literal
     ;
 
 // Math
-expr : expr_1 ;
-expr_1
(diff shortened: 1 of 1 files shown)


falkecarlsen:
I think you should've written a short list of what has been updated instead of having the reviewer deduce it from the changed code. Omitting this also makes it impossible to catch things that may have been forgotten to be updated.

Generally, make a short and concise description of the logical changes contained in the given PR. Please do this for all future PRs.

parser/cellmata.g4
```diff
@@ -165,8 +161,9 @@ PAREN_START : '(' ;
 PAREN_END : ')' ;
 END : ';' ;
 
-OP_COMPARE : 'is' ;
-OP_NOT : 'not' ;
+OP_COMPARE : '==' ;
+OP_COMPARE_not : '!=' ;
```
falkecarlsen: Whoops, seems like the token `OP_COMPARE_not`, note the lower case, does not match the applied usage at line 96. Please fix.

falkecarlsen (changes requested):
One issue with mismatch of token and application of this.

parser/cellmata.g4
jenrik: Typo has been corrected in 8a667a6db564bfe5beceb4fe20afe51db317beab

falkecarlsen (approved):

PR #13, opened 2019-03-13
Added function calls

jenrik:
Changed function calls to accept arbitrary functions.

### DIFF
--- a/parser/cellmata.g4
+++ b/parser/cellmata.g4
@@ -2,7 +2,7 @@ grammar cellmata;
 
 start : world_dcl body EOF;
 
-body : (state_decl | neighbourhood_decl | const_decl)*;
+body : (state_decl | neighbourhood_decl | const_decl | func_decl)*;
 const_decl : STMT_CONST const_ident ASSIGN expr ;
 const_ident : IDENT ;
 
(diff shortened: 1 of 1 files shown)


parser/cellmata.g4
```diff
@@ -143,11 +144,11 @@ expr_11
...
 
-// Built-in funcitons
-func : (func_count | func_rand | func_abs) ;
-func_count : FUNC_COUNT PAREN_START neighbourhood_ident LIST_SEP state_ident PAREN_END ;
-func_rand : FUNC_RAND PAREN_START DIGITS PAREN_END ;
-func_abs : FUNC_ABS PAREN_START expr PAREN_END ;
+// Functions
+func : func_ident PAREN_START (expr (LIST_SEP expr)* )? PAREN_END ;
+func_ident : IDENT ;
+func_decl : STMT_FUNC func_ident PAREN_START func_decl_arg (LIST_SEP func_decl_arg)* PAREN_END type_ident code_block ;
```
falkecarlsen, line 150: No issues, I'm just interested to see how type declaration of a function **after** parameter-declaration will turn out looking.

falkecarlsen (approved):

parser/cellmata.g4
jenrik: For reference it's probably gonna look very similar to Kotlin and Go

PR #14, opened 2019-03-13
Added modulo operator

jenrik:


parser/cellmata.g4
```diff
@@ -116,6 +116,7 @@ expr_6
 expr_7
     : expr_7 OP_MULTIPLY expr_8 # multiplictionExpr
     | expr_7 OP_DIVIDE expr_8 # divisionExpr
+    | expr_7 OP_MODULUS expr_8 # modulusExpr
```
falkecarlsen:
```suggestion
    | expr_7 OP_MODULO expr_8 # modulusExpr
```

falkecarlsen (changes requested):
The correct spelling is "modulo", both in Danish and English, and the syntax should reflect that. Note that line 177 needs to reflect this change as well: `OP_MODULUS : '%' ; // PRAISE!`

Maybe we should remove the in-joke while we're at it. No hate :innocent:.

falkecarlsen (approved):

PR #16, opened 2019-03-14
Moved compiler into this repository

jenrik:
Moved the experimental Kotlin compiler into this repository, and the ANTLR grammar into the compiler directory.


compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -0,0 +1,147 @@
...
+ * If you see this string in running code it means that the variable hasn't been initialized.
+  */
+const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
+
+sealed class AST
+
+data class RootNode(val world: WorldNode, val body: List<Decl>) : AST()
+
+/*
+* World definition
```
falkecarlsen: We should distributed a codestyle-config for IntelliJ, which the group agrees upon.

jenrik: I agree

PR #20, opened 2019-03-14
Removed tokens for built in functions

### LINKED ISSUE #19: Built-in function tokens grabs function-ident
These should be removed.

falkecarlsen:
This resolves #19, as these tokens are no longer grabbed by tokens and unable to be matched as an ident.


NicEastvillage (approved):
Nice

jenrik (approved):

PR #23, opened 2019-03-14
Implemented support for state and neighbourhoods as types

falkecarlsen:
Implemented support for state and neighbourhoods as types to function parameters, by modifying `type_spec` production to include the `STMT_*` tokens for state and neighbourhoods.

Also fixed minor typo: `tyopeNumber`.

### DIFF
--- a/compiler/src/main/antlr/Cellmata.g4
+++ b/compiler/src/main/antlr/Cellmata.g4
@@ -62,7 +62,9 @@ type_ident : IDENT | type_spec ;
 type_spec
     : array_decl # typeArray
     | TYPE_BOOLEAN # typeBoolean
-    | TYPE_NUMBER # tyopeNumber
+    | TYPE_NUMBER # typeNumber
+    | STMT_NEIGHBOUR # typeNeighbour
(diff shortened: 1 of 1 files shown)


compiler/src/main/antlr/Cellmata.g4
```diff
@@ -62,7 +62,9 @@ type_ident : IDENT | type_spec ;
 type_spec
     : array_decl # typeArray
     | TYPE_BOOLEAN # typeBoolean
-    | TYPE_NUMBER # tyopeNumber
+    | TYPE_NUMBER # typeNumber
+    | STMT_NEIGHBOUR # typeNeighbour
```
jenrik, line 66: It a bit weird that we have `TYPE_` prefix on some but `STMT_` prefix on other, but that is probably unfixable.

jenrik (approved):

compiler/src/main/antlr/Cellmata.g4
falkecarlsen: Agreed, maybe write an issue on it?

PR #27, opened 2019-03-16
Collapsed expression productions

jenrik:
Having individual productions for each predence level is unneccesary
since ANTLR 4 does predence based on declaration order. This change
shouldn't affect predence of any operators.
Note ANTLR 4 also does left-recursion elimination for direct
left-recursion, so that isn't an issue.


falkecarlsen (approved):
Looks good, and is infinitely more readable than previous iteration. 
I also wrote #29, which I think we're missing from expression grammar, but this is not relevant to the purpose of this PR.

PR #28, opened 2019-03-17
Added multi states

jenrik:
Added support for repeating states that share logic.
Added a new literal that can be used to reference the variant of the current current cell if it's in a multi state.

### DIFF
--- a/compiler/src/main/antlr/Cellmata.g4
+++ b/compiler/src/main/antlr/Cellmata.g4
@@ -24,7 +24,7 @@ world_cellsize
 world_cellsize_value : integer_literal ;
 
 // State
-state_decl : STMT_STATE state_ident state_rgb code_block ;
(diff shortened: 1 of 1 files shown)


falkecarlsen (changes requested):
An example of a state declaration according to this grammar:

```
state foostate [5] (255, 255, 255) {
  become foofunc(rand(#)); //assuming that foofunc returns a state[n] depending on state-number
  // become foostate[4]; - this would be legal
}
```

I like that the square brackets distinguishes between RGB-declaration and multi-state-declaration. It kinda looks like an array-size declaration, which is somewhat relevant to multi-states.

~~Are we still allowing simply saying `become foostate;` and then implicitly passing the current state, meaning that a foostate, where `# == 5` will, when `become foostate;` become a foostate where `# == 5`? Should we implicitly increment the state?~~

I think we should discuss these issues with the group before merging this PR.

NicEastvillage (approved):

falkecarlsen (approved):

PR #31, opened 2019-03-18
Completed preliminary parse tree to AST mapper

jenrik:
Added terminator to constant declaration.
Named production parts for ease of use.
Collapsed expressions. ANTLR 4 automaticly does precedence based on
declaration order.
Updated mapper to use named production parts.
Added missing modulo expression mapping.
Added missing function expression mapping.
Added state index expression mapping.
Added missing array literal expression mapping.
Added missing if statement mapping.
Added missing return statement mapping.
Added missing neighbourhood declaration mapping.
Added function declaration mapping.
Added syntax stress example.

### DIFF
--- a/compiler/src/main/antlr/Cellmata.g4
+++ b/compiler/src/main/antlr/Cellmata.g4
@@ -7,8 +7,6 @@ const_decl : STMT_CONST const_ident ASSIGN expr END ;
 const_ident : IDENT ;
 
 // World
-world_dcl : STMT_WORLD BLOCK_START world_size world_tickrate? world_cellsize? BLOCK_END ;
-world_size_dim : integer_literal SQ_BRACKET_START world_size_dim_finite SQ_BRACKET_END # dimFinite ;
 world_dcl : STMT_WORLD BLOCK_START size=world_size tickrate=world_tickrate? cellsize=world_cellsize? BLOCK_END ;
 world_size : WORLD_SIZE ASSIGN width=world_size_dim (LIST_SEP height=world_size_dim)?;
 world_size_dim : integer_literal SQ_BRACKET_START world_size_dim_finite SQ_BRACKET_END # dimFinite ;
@@ -28,11 +26,11 @@ world_cellsize_value : integer_literal ;
 // State
 state_decl : STMT_STATE state_ident (SQ_BRACKET_START integer_literal SQ_BRACKET_END)? state_rgb code_block ;
 state_ident : IDENT ;
-state_rgb : PAREN_START integer_literal LIST_SEP integer_literal LIST_SEP integer_literal PAREN_END ;
+state_rgb : PAREN_START red=integer_literal LIST_SEP green=integer_literal LIST_SEP blue=integer_literal PAREN_END ;
 
 // Code
 code_block : BLOCK_START stmt* BLOCK_END ;
-stmt : (if_stmt | become_stmt | assign_stmt | increment_stmt | decrement_stmt | return_stmt | END) ;
+stmt : (if_stmt | become_stmt | assign_stmt | increment_stmt | decrement_stmt | return_stmt) ;
 
 assign_stmt : STMT_LET? (var_ident | array_lookup) ASSIGN expr END ;
 if_stmt : if_stmt_if if_stmt_elif* if_stmt_else? ;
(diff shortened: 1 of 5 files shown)


compiler/src/main/antlr/Cellmata.g4
```diff
@@ -93,37 +100,37 @@ bool_literal
...
+    | left=expr OP_DIVIDE right=expr # divisionExpr
+    | left=expr OP_MODULO right=expr # moduloExpr
+    | OP_INCREMENT value=expr # preIncExpr
+    | OP_DECREMENT value=expr # preDecExpr
+    | OP_PLUS value=expr # positiveExpr
+    | OP_MINUS value=expr # negativeExpr
+    | OP_NOT value=expr # inverseExpr
+    | value=expr OP_INCREMENT # postIncExpr
+    | value=expr OP_DECREMENT # postDecExpr
+    | value=expr SQ_BRACKET_START DIGITS SQ_BRACKET_END # arrayLookupExpr
```
NicEastvillage:
```suggestion
    | value=expr SQ_BRACKET_START expr SQ_BRACKET_END # arrayLookupExpr
```

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -76,27 +75,22 @@ data class InverseExpr(val ctx: ParseTree, val value: Expr) : Expr()
 
 data class ArrayLookupExpr(val ctx: ParseTree, val ident: String = MAGIC_UNDEFINED_STRING, val index: Int = -1) : Expr()
 
-data class ArrayBodyExpr(val ctx: ParseTree, val type: String?, val values: List<Expr>) : Expr()
+data class ArrayBodyExpr(val ctx: ParseTree, val type: String? = MAGIC_UNDEFINED_STRING, val values: List<Expr>) : Expr()
```
NicEastvillage: I think we should make types to an enum.

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -114,14 +108,23 @@ data class StateDecl(
     val body: List<Stmt>
 ) : Decl()
 
-data class Coordinate(val axes: List<Int>)
+data class Coordinate(val ctx: ParseTree, val axes: List<Int> = emptyList())
```
NicEastvillage, line 114:
```suggestion
data class Coordinate(val ctx: ParseTree, val coordX: Int, val coordY: Int?)
```

compiler/src/main/kotlin/ast/mapper.kt
```diff
@@ -109,34 +114,55 @@ private fun visitExpr(node: ParseTree): Expr {
...
+                        expr = visitExpr(it.if_stmt_block().if_stmt_condition().expr()),
+                        block = visitCodeBlock(it.if_stmt_block().code_block())
+                    )
+                }
+            ).flatten(),
+            elseBlock = when (node.if_stmt_else()) {
+                // null???
+                is CellmataParser.If_stmt_elseContext -> visitCodeBlock(node.if_stmt_else().code_block())
+                else -> throw AssertionError("Unexpected tree node")
+            }
```
NicEastvillage, line 163: Here will be problems when the else-case is not present as the comment `// nul???` also suggests

compiler/src/main/kotlin/ast/mapper.kt
```diff
@@ -146,38 +172,37 @@ private fun visitDecl(node: ParseTree): Decl {
...
     }
 }
 
+fun visitCodeBlock(block: CellmataParser.Code_blockContext): List<Stmt> {
+    return block.children
+        .filter { it !is TerminalNode } // Remove terminals
+        .filter { !(it is CellmataParser.StmtContext && it.childCount == 0) }
+        .filter { !(it is CellmataParser.StmtContext && it.getChild(0) is TerminalNode ) }
+        .map { visitStmt(it) }
+}
```
NicEastvillage, line 202: This is weird. What is this filtering for? Why isn't children just 0 or more statements?

compiler/src/main/kotlin/main.kt
```diff
@@ -22,4 +16,6 @@ fun main() {
     val ast = visit(startContext)
 
     println(ast)
+
+    listOf<Unit>()
```
NicEastvillage: Why??

NicEastvillage (changes requested):
A few things seems a bit weird. Check those please.

compiler/src/main/kotlin/ast/ast.kt
jenrik: Having the axes as a list allows for easier data manipulation and validation, and have proven useful to me

compiler/src/main/kotlin/ast/ast.kt
jenrik: I'm not sure enums are the right solution, how would you represent an array of integers?

jenrik:
The ability to use a semicolon as a no-op statement has been removed

NicEastvillage (approved):

PR #34, opened 2019-03-19
Added symbol table

jenrik:
Added AST visitor, and base implementation of the visitor, see 
a71d6f0.
Added a AST walker that extract values from the parse tree, see 0059a2e.
Added a symbol table.
Added symbol walker that extract symbols and does declaration-before-use checking, 
d75818d.

### DIFF
--- a/compiler/src/main/kotlin/ast/symbol.kt
+++ b/compiler/src/main/kotlin/ast/symbol.kt
@@ -1,0 +1,53 @@
+package dk.aau.cs.d409f19.cellumata.ast
+
+import java.util.*
+
+data class Table(
+    val symbols: MutableMap<String, AST> = mutableMapOf(),
+    val tables: MutableList<Table> = mutableListOf()
+)
+
+class SymbolRedefinitionException(val ident: String): Exception("\"$ident\" is already defined")
+
+class SymbolTable {
+    private val root: Table = Table()
+    private val scopeStack: Stack<Table> = Stack()
+
+    init {
+        scopeStack.push(root)
+    }
+
+    fun insertSymbol(ident: String, node: AST) {
+        val table = scopeStack.peek()
+
+        if(table.symbols.containsKey(ident)) {
+            throw SymbolRedefinitionException(ident = ident)
+        }
+
+        table.symbols[ident] = node
+    }
+
+    fun getSymbol(name: String): AST? {
+        for (table in scopeStack) {
+            if (table.symbols.containsKey(name)) {
+                return table.symbols.get(name)
+            }
+        }
+        return null
+    }
+
+    fun createScope() {
+        val newScope = Table()
+        scopeStack.peek().tables.add(newScope)
+        scopeStack.push(newScope)
+    }
+
+    fun closeScope() {
+        assert(scopeStack.size > 1) { "Tried to remove root scope from symbol table" }
+        scopeStack.pop()
+    }
+
+    fun openScope(index: Int) {
+        scopeStack.push(scopeStack.peek().tables[index])
+    }
+}
--- a/compiler/src/main/kotlin/ast/symbol.kt
+++ b/compiler/src/main/kotlin/ast/symbol.kt
@@ -7,7 +7,32 @@ data class Table(
     val tables: MutableList<Table> = mutableListOf()
 )
 
-class SymbolRedefinitionException(val ident: String): Exception("\"$ident\" is already defined")
+class SymbolRedefinitionException(val ident: String) : Exception("\"$ident\" is already defined")
+
+private val RESERVED_SYMBOLS: List<String> = listOf(
+    "world",
+    "neighbourhood",
+    "state",
+    "function",
+    "int",
+    "float",
+    "bool",
+    "become",
+    "return",
+    "if",
+    "elif",
+    "else",
+    "let",
+    "for",
+    "continue",
+    "break",
+    "rand",
+    "abs",
+    "floor",
+    "ceil",
+    "sqrt",
+    "pow"
+)
 
 class SymbolTable {
     private val root: Table = Table()
@@ -20,7 +45,11 @@ class SymbolTable {
     fun insertSymbol(ident: String, node: AST) {
         val table = scopeStack.peek()
 
-        if(table.symbols.containsKey(ident)) {
+        if (RESERVED_SYMBOLS.contains(ident)) {
+            throw SymbolRedefinitionException(ident = ident)
+        }
+
+        if (table.symbols.containsKey(ident)) {
             throw SymbolRedefinitionException(ident = ident)
         }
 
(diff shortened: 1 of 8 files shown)


jenrik:
The symbol table now rejects attempts to insert symbols that matches language keywords, and builtin functions

NicEastvillage:
`.circleci/config.yml` does not seem relevant for this change.

compiler/src/main/kotlin/ast/symbol.kt
```diff
@@ -0,0 +1,82 @@
...
+    "let",
+    "for",
+    "continue",
+    "break",
+    "rand",
+    "abs",
+    "floor",
+    "ceil",
+    "sqrt",
+    "pow"
```
NicEastvillage, line 34: pow, sqrt, ceil og andre indbyggede funktioner er ikke reserverede nøgleord. Disse funktioner skal lægges i symbol table før brug. Der skal muligvis et eller andet hacky til for at lave en AST node til dem. Umiddelbart tænker jeg vi kan lave en unik AST node til hver af dem.

jenrik:
the test and CircleCI related commits has been removed

compiler/src/main/kotlin/walkers/value.kt
```diff
@@ -0,0 +1,133 @@
...
+class BoolParsingException(val value: String, val ctx: BoolLiteral): ParsingException(msg = "\"$value\"\"$value\" is not a valid boolean")
+
+class FloatParsingException(val value: String, val ctx: FloatLiteral): ParsingException(msg = "\"$value\" is not a valid float")
+
+class ColorParsingException(val value: String, val ctx: StateDecl): ParsingException(msg = "\"$value\" is not a valid color")
+
+class CoordinateParsingException: ParsingException("Inconsistent amount axes in neighbourhood")
+
+
+class ParseTreeValueWalker : BaseASTVisitor() {
```
NicEastvillage: Jeg ville kalde denne for LiteralExtractorVisitor

compiler/src/main/kotlin/walkers/symbol.kt
```diff
@@ -0,0 +1,60 @@
+package dk.aau.cs.d409f19.cellumata.walkers
+
+import dk.aau.cs.d409f19.cellumata.ast.*
+import java.lang.Exception
+
+class SymbolException(val ident: String) : Exception("\"$ident\" was used before it was declared")
+
+class SymbolWalker(val symbolTable: SymbolTable) : BaseASTVisitor() {
```
NicEastvillage: Dette ligner ScopeCheckVisitor.

NicEastvillage (dismissed):
Jeg har nogle ændringer til hvordan indbyggede funktioner gemmes i symbol table (dette kan vi ændre i et nyt PR), og nogle forslag til navneændringer.  Men ellers ser indholdet korrekt ud.

NicEastvillage:
Btw, make sure to negative numbers are extracted as negative values in LiteralExtractorVisitor (aka ParseTreeValueWalker). PR #41 will change the definition of integer_literals and float_literals.

compiler/src/main/kotlin/ast/symbol.kt
jenrik: I think we should take care of this in a future pull request since resolving it is a rather complex matter involving how we should handle function overloading.

falkecarlsen: In that case, unless @NicEastvillage disagrees, we should make an issue to track the work.

jenrik:
The conversion from number literals to there respective value is done through kotlin's `String.toInt()` and `String.toFloat()` that includes negative number. So if the integer and float tokens consume the minus kotlin will interpret them as negative.

falkecarlsen (dismissed):
Looks good, all relevant comments have been resolved.

Please write an issue regarding the built-in functions symbols, though.

compiler/src/main/kotlin/walkers/symbol.kt
```diff
@@ -0,0 +1,60 @@
...
+            symbolTable.insertSymbol(node.ident, node)
+        }
+    }
+
+    override fun visit(node: ConditionalBlock) {
+        symbolTable.createScope()
+        super.visit(node)
+        symbolTable.closeScope()
+    }
+}
```
NicEastvillage, line 68: You have VarExpr where symbol declaration is checked, but FuncExpr, StateExpr, NeighbourhoodExpr (or whatever their names are) are not checked if the name is already declared.

NicEastvillage (changes requested):

compiler/src/main/kotlin/walkers/symbol.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    override fun visit(node: ConstDecl) {
+        super.visit(node)
+        symbolTable.insertSymbol(node.ident, node)
+    }
+
+    override fun visit(node: StateDecl) {
+        symbolTable.createScope()
```
falkecarlsen, line 35: Does not check if the color declaration contain undeclared constants.

compiler/src/main/kotlin/walkers/symbol.kt
jenrik: State names and neighbourhood names are currently matched by VarExpr, so maybe we should rename VarExpr to something more generic.

jenrik: e662dc1 implements code that checks functions are declared before they are used.

NicEastvillage: I suggest `Identifier` for named expressions. Or `NamedExpression`.

NicEastvillage: I will resolve and approve after rename.

falkecarlsen (changes requested):
Another missing functionality, can be put off, if necessary.

compiler/src/main/kotlin/walkers/symbol.kt
jenrik: I have renamed `VarExpr` to `NamedExpr` in 8247d7f

compiler/src/main/kotlin/walkers/symbol.kt
jenrik: I think we should discus what expressions are allowed in state colors.

NicEastvillage (approved):

compiler/src/main/kotlin/walkers/symbol.kt
jenrik: I looked into the grammar and found that colors currently are limited to integer literals, so we should fill issue regarding expression as colors and the limitations we probably gonna have to put on them.

falkecarlsen (approved):

PR #36, opened 2019-03-19
Added classes/objects that describes types and some helper functions for type conversion

### LINKED ISSUE #32: Use a sealed class for types instead of strings
Current types are stored as strings, but sinces we have a finite number of types enums would be better.

NicEastvillage:
Resolves #32


jenrik:
I think we should hold of with merging this in to master until we have done some practical testing of the design

falkecarlsen (approved):

PR #45, opened 2019-03-21
Added mapping for for-loop, break, and continue statements

### LINKED ISSUE #43: Missing AST node(s) for for-loops

### LINKED ISSUE #44: Missing `continue` and `break` statements

NicEastvillage:
I had to remove `expr` as valid post-iteration grammar, since we can only have either expressions or statements, so I chose the strict option of only allowing assign statements.

Resolves #44 and resolves #43

### DIFF
--- a/compiler/src/main/antlr/Cellmata.g4
+++ b/compiler/src/main/antlr/Cellmata.g4
@@ -52,10 +52,10 @@ decrement_stmt
 return_stmt : STMT_RETURN expr END ;
 
 //For-loop
-for_stmt : STMT_FOR PAREN_START for_variable? END for_condition END for_increment? PAREN_END code_block ;
-for_variable : assignment ;
+for_stmt : STMT_FOR PAREN_START for_init? END for_condition END for_post_iteration? PAREN_END code_block ;
+for_init : assignment ;
 for_condition : expr ;
-for_increment : expr | assignment ;
+for_post_iteration : assignment ;
 
 // Neighbourhood
 neighbourhood_decl : STMT_NEIGHBOUR neighbourhood_ident neighbourhood_code ;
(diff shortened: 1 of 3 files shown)


compiler/src/main/kotlin/ast/mapper.kt
```diff
@@ -1,6 +1,7 @@
 package dk.aau.cs.d409f19.cellumata.ast
 
 import cs.aau.dk.d409f19.antlr.CellmataParser
+import javafx.scene.control.Cell
```
falkecarlsen: Hmm, this seems unused. Please remove if it is.

falkecarlsen (changes requested):
Looks good, a stray import made it into the PR though.

compiler/src/main/antlr/Cellmata.g4
```diff
@@ -52,10 +52,12 @@ decrement_stmt
 return_stmt : STMT_RETURN expr END ;
 
 //For-loop
-for_stmt : STMT_FOR PAREN_START for_variable? END for_condition END for_increment? PAREN_END code_block ;
-for_variable : assignment ;
+for_stmt : STMT_FOR PAREN_START for_init? END for_condition END for_post_iteration? PAREN_END code_block ;
+for_init : assignment ;
 for_condition : expr ;
-for_increment : expr | assignment ;
+for_post_iteration : assignment ;
```
MathiasMehl, line 58: if expr is omitted here, increment (i.e i-- or i++) is not possible in for loop as an update

NicEastvillage: That is correct. We are removing unary increment and decrement soon.

falkecarlsen (approved):

PR #53, opened 2019-03-26
Naming of CST to AST reduce-functions

falkecarlsen:
Instead of using "visit" to denote how the *trimming* of the CST is done, "reduce" is used instead. Furthermore, instead of calling the source-file `mapper.kt`, it is now called `reduce.kt`


compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -4,190 +4,190 @@ import dk.aau.cs.d409f19.antlr.CellmataParser
...
         )
-        is CellmataParser.NumberLiteralContext -> visitExpr(node.value)
-        is CellmataParser.BoolLiteralContext -> visitExpr(node.value)
+        is CellmataParser.NumberLiteralContext -> reduceExpr(node.value)
+        is CellmataParser.BoolLiteralContext -> reduceExpr(node.value)
         is CellmataParser.Bool_literalContext -> BoolLiteral(ctx = node)
-        is CellmataParser.Number_literalContext -> visitExpr(node.getChild(0))
-        is CellmataParser.IntegerLiteralContext -> visitExpr(node.value)
-        is CellmataParser.FloatLiteralContext -> visitExpr(node.value)
+        is CellmataParser.Number_literalContext -> reduceExpr(node.getChild(0)) // ToDo is this correct?
```
jenrik:
```suggestion
        is CellmataParser.Number_literalContext -> reduceExpr(node.getChild(0))
```

jenrik: I previously removed this comment because i determined `node.getChilde(0)` to be correct

jenrik (changes requested):

jenrik (approved):

PR #54, opened 2019-03-26
Simple error system for handling minor and major errors and printing details to console.

### LINKED ISSUE #47: Better error system that doesn't abort on first error

NicEastvillage:
The ErrorLogger object holds all errors so far.

Small compile errors derive from the CompileError interface (or more likely the ErrorFromContext that uses an ANTLR's parser context). CompileErrors should be registered in the ErrorLogger and compilation should continue. After each phase in the compiler, the ErrorLogger is check for errors and prints any errors accumulated, stopping the compilation.

Major compile errors should use throw the TerminatedCompilationException which will stop the compilation immediately.

Any other exception is considered an error in the compiler, i.e. an error by us and not the programmer of the .cell program.

Resolves #47

### DIFF
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -11,7 +11,7 @@ const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
 
 sealed class AST(open val ctx: ParserRuleContext)
 
-data class RootNode constructor(
+data class RootNode(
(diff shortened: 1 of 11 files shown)


falkecarlsen (dismissed):
Looks very nice!

falkecarlsen:
I committed some fixes in https://github.com/d409f19/cellmata/pull/54/commits/7fbf70c16857e7fa784b2686c96139882723bb84. Hope you like.

falkecarlsen (dismissed):

falkecarlsen (approved):

PR #56, opened 2019-03-26
Implemented type checker

jenrik:
Support for implicit conversions are missing.
Added float and integer tokens to grammar.
Changed the TypedNode interface to allow for more classes to contain a type, and to have a generic way of accessing that type.
Move type of literals to their class definition.
Deprecated SymbolTable in favor of the new CreatingSymbolTableSession.
Removed typeFromString in favor of typeFromCtx.
Added ScopedASTVisitor to traverse the AST while automaticly opening and closing scopes.
Updated test to include ParseTreeValueWalker, SymbolWalker, and TypeChecker.
Changed type names to use the new short version.

### DIFF
--- a/compiler/src/main/kotlin/ast/symbol.kt
+++ b/compiler/src/main/kotlin/ast/symbol.kt
@@ -34,6 +34,7 @@ private val RESERVED_SYMBOLS: List<String> = listOf(
     "pow"
 )
 
+@Deprecated("Use CreatingSymbolTableSession")
 class SymbolTable {
     private val root: Table = Table()
     private val scopeStack: Stack<Table> = Stack()
@@ -65,6 +66,18 @@ class SymbolTable {
         return null
     }
 
+    fun getSymbolType(name: String): Type? {
+        val symbol = getSymbol(name) ?: return null
+
+        return when(symbol) {
+            is TypedNode -> symbol.getType()
+            is StateDecl -> StateType
+            is ConstDecl -> symbol.type
+            is FuncDecl -> symbol.returnType
+            else -> null
+        }
+    }
+
     fun createScope() {
         val newScope = Table()
         scopeStack.peek().tables.add(newScope)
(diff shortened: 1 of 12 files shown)


compiler/src/main/kotlin/ast/symbol.kt
```diff
@@ -79,4 +92,111 @@ class SymbolTable {
...
+
+class CreatingSymbolTableSession(symbolTable: Table) {
+    private val scopeStack: Stack<Table> = Stack()
+
+    init {
+        scopeStack.push(Table(tables = mutableListOf(symbolTable))) // God scope
+        scopeStack.push(symbolTable)
+    }
+
+    fun openScope() {
```
NicEastvillage, line 105: I suggest renaming `openScope()` to `createScope()` to make it clear that a new scope is created, and it is not the next scope that is opened like in the `ViewingSymbolTableSession`. I.e. this will make the difference between creating and viewing scopes in the symbol table more clear.

compiler/src/main/kotlin/walkers/value.kt
```diff
@@ -83,10 +83,10 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
     override fun visit(node: FuncDecl) {
         super.visit(node)
 
-        node.args = node.ctx.func_decl_arg().map { FunctionArgs(it.IDENT().text, it.type_ident().text) }.toList()
+        node.args = node.ctx.func_decl_arg().map { FunctionArgs(it.IDENT().text, typeFromCtx(it.type_ident())) }.toList()
```
NicEastvillage, line 86: TypeChecking in the LiteralValueExtractor? Tell me why this is done here?

compiler/src/main/kotlin/walkers/value.kt
```diff
@@ -83,10 +83,10 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
...
 
         node.ident = node.ctx.func_ident().text
-        node.returnType = node.ctx.type_ident().text
+        node.returnType = typeFromCtx(node.ctx.type_ident())
```
NicEastvillage, line 89: More TypeChecking of functions in the LiteralValueExtractor.

NicEastvillage (changes requested):
A few things here and there, BUT documentation is lacking almost everywhere.

jenrik:
This pull request has been accidentally merged while merging #57, I will check the comments and update this PR accordingly.

PR #57, opened 2019-03-26
Documented source code

jenrik:
This pull request depends on #56

### DIFF
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -10,9 +10,9 @@ import org.antlr.v4.runtime.tree.ParseTree
 const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
 
 /**
- * This classes is the basis for all nodes in the abstract syntax tree.
+ * This class is the basis for all nodes in the abstract syntax tree.
  *
- * It's sealed such that it's possible to use Kotlin's when statement to without an else.
+ * It's sealed such that it's possible to use Kotlin's when statement without an else.
  */
 sealed class AST
 
@@ -307,7 +307,7 @@ data class ContinueStmt(val ctx: CellmataParser.Continue_stmtContext) : Stmt()
 
 /**
  * Represents a become statement.
- * The become statements terminates a state block, and changes the the state of the cell being evaluated.
+ * The become statements terminates a state block, and changes the state of the cell being evaluated.
  */
 data class BecomeStmt(val ctx: CellmataParser.Become_stmtContext, val state: Expr) : Stmt()
 
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -9,19 +9,42 @@ import org.antlr.v4.runtime.tree.ParseTree
  */
 const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
 
+/**
+ * This classes is the basis for all nodes in the abstract syntax tree.
+ *
+ * It's sealed such that it's possible to use Kotlin's when statement to without an else.
+ */
 sealed class AST
 
+/**
+ * RootNode represents a .cell file
+ */
 data class RootNode(val world: WorldNode, val body: List<Decl>) : AST()
 
 /*
  * World definition
  */
+/**
+ * The types of world bounds that are supported.
+ */
 enum class WorldType {
+    /**
+     * WRAPPING indicate that if a CA tries read beyond the world limit its lookup will automatically be from the other side of the world.
+     */
     WRAPPING,
+    /**
+     * EDGE indicates that all cells beyond the world limit has a fixed state.
+     */
     EDGE,
+    /**
+     * UNDEFINED is a compiler internal value used to indicate that no type has been parsed yet.
+     */
     UNDEFINED
 }
 
+/**
+ * Nodes that hold a value or that can produce a value should implement TypedNode to expose the type of that value
+ */
 interface TypedNode {
     fun getType(): Type?
     fun setType(type: Type?)
@@ -29,6 +52,9 @@ interface TypedNode {
 
 data class WorldDimension(val size: Int = -1, val type: WorldType = WorldType.UNDEFINED)
 
+/**
+ * WorldNode represent the world definition.
+ */
 data class WorldNode(
     val ctx: ParseTree,
     val dimensions: List<WorldDimension> = emptyList(),
@@ -38,6 +64,9 @@ data class WorldNode(
 
 /*
  * Expressions
+ */
+/**
+ * An expression is either a literal, or some operation that produces a value.
  */
 abstract class Expr(type: Type? = UncheckedType) : AST(), TypedNode {
     private var atype: Type?
@@ -93,9 +122,17 @@ data class NegativeExpr(val ctx: CellmataParser.NegativeExprContext, val value: 
 
 data class InverseExpr(val ctx: CellmataParser.InverseExprContext, val value: Expr) : Expr()
 
-data class ArrayLookupExpr(val ctx: CellmataParser.ArrayLookupExprContext, val ident: String = MAGIC_UNDEFINED_STRING, val index: Expr) : Expr()
+data class ArrayLookupExpr(
+    val ctx: CellmataParser.ArrayLookupExprContext,
+    val ident: String = MAGIC_UNDEFINED_STRING,
+    val index: Expr
+) : Expr()
 
-data class ArrayBodyExpr(val ctx: CellmataParser.ArrayValueExprContext, val values: List<Expr>, val declaredType: Type? = UncheckedType) : Expr()
+data class ArrayBodyExpr(
+    val ctx: CellmataParser.ArrayValueExprContext,
+    val values: List<Expr>,
+    val declaredType: Type? = UncheckedType
+) : Expr()
 
 data class ParenExpr(val ctx: CellmataParser.ParenExprContext, val expr: Expr) : Expr()
 
@@ -103,27 +140,55 @@ data class NamedExpr(val ctx: ParseTree, var ident: String = MAGIC_UNDEFINED_STR
 
 data class ModuloExpr(val ctx: CellmataParser.ModuloExprContext, val left: Expr, val right: Expr) : Expr()
 
-data class FuncExpr(val ctx: CellmataParser.FuncExprContext, val args: List<Expr>, var ident: String = MAGIC_UNDEFINED_STRING) : Expr()
+/**
+ * Presents a function call in abstract syntax tree
+ */
+data class FuncExpr(
+    val ctx: CellmataParser.FuncExprContext,
+    val args: List<Expr>,
+    var ident: String = MAGIC_UNDEFINED_STRING
+) : Expr()
 
 data class StateIndexExpr(val ctx: CellmataParser.StateIndexExprContext) : Expr(type = IntegerType)
 
 // Literals
+/**
+ * Represents a concrete integer value in the abstract syntax tree
+ */
 data class IntLiteral(val ctx: CellmataParser.Integer_literalContext, var value: Int = -1) : Expr(type = IntegerType)
 
+/**
+ * Represents a concrete boolean value in the abstract syntax tree
+ */
 data class BoolLiteral(val ctx: CellmataParser.Bool_literalContext, var value: Boolean = false) : Expr(type = BooleanType)
 
-data class FloatLiteral(val ctx: CellmataParser.Float_literalContext, var value: Float = 0.0F): Expr(type = FloatType)
+/**
+ * Represent a concrete float value in the abstract syntax tree
+ */
+data class FloatLiteral(val ctx: CellmataParser.Float_literalContext, var value: Float = 0.0F) : Expr(type = FloatType)
 
 // Type conversion
+/**
+ * Internal node used to present an implicit conversion from integer to float.
+ */
 data class IntToFloatConversion(val expr: Expr) : Expr()
 
+/**
+ * Internal node used to represent an implicit conversion from a array of states to a neighbourhood.
+ */
 data class StateArrayToActualNeighbourhoodConversion(val expr: Expr) : Expr()
 
 /*
  * Declarations
  */
+/**
+ * Represents global declarations like states, constants and functions
+ */
 sealed class Decl : AST() // state, const, func
 
+/**
+ * Represents constants declarations
+ */
 data class ConstDecl(
     val ctx: CellmataParser.Const_declContext,
     var ident: String = MAGIC_UNDEFINED_STRING,
@@ -131,6 +196,9 @@ data class ConstDecl(
     var type: Type? = UncheckedType
 ) : Decl()
 
+/**
+ * Represent a state declaration
+ */
 data class StateDecl(
     val ctx: CellmataParser.State_declContext,
     var ident: String = MAGIC_UNDEFINED_STRING,
@@ -140,15 +208,26 @@ data class StateDecl(
     val body: List<Stmt>
 ) : Decl()
 
+/**
+ * Represents a single point in a neighbourhood declaration
+ */
 data class Coordinate(val ctx: ParseTree, val axes: List<Int> = emptyList())
 
+/**
+ * Represents a neighbourhood declaration
+ */
 data class NeighbourhoodDecl(
     val ctx: CellmataParser.Neighbourhood_declContext,
     var ident: String = MAGIC_UNDEFINED_STRING,
     var coords: List<Coordinate> = emptyList()
 ) : Decl()
 
-data class FunctionArgs(val ident: String, private var type: Type?): AST(), TypedNode {
+/**
+ * Represent a single argument in a function declaration
+ *
+ * @see FuncDecl
+ */
+data class FunctionArgs(val ident: String, private var type: Type?) : AST(), TypedNode {
     override fun getType(): Type? {
         return type
     }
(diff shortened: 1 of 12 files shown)


compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -9,21 +9,52 @@ import org.antlr.v4.runtime.tree.ParseTree
  */
 const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
 
+/**
+ * This classes is the basis for all nodes in the abstract syntax tree.
+ *
+ * It's sealed such that it's possible to use Kotlin's when statement to without an else.
```
falkecarlsen: I think an extra "to" made it's way in
```suggestion
 * It's sealed such that it's possible to use Kotlin's when statement without an else.
```

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -9,21 +9,52 @@ import org.antlr.v4.runtime.tree.ParseTree
  */
 const val MAGIC_UNDEFINED_STRING = "<<THIS IS A MAGIC STRING, UNDEFINED>>"
 
+/**
+ * This classes is the basis for all nodes in the abstract syntax tree.
```
falkecarlsen:
```suggestion
 * This class is the basis for all nodes in the abstract syntax tree.
```

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -74,39 +122,88 @@ data class NegativeExpr(val ctx: CellmataParser.NegativeExprContext, val value:
...
 
-data class StateIndexExpr(val ctx: CellmataParser.StateIndexExprContext) : Expr()
+data class StateIndexExpr(val ctx: CellmataParser.StateIndexExprContext) : Expr(type = IntegerType)
 
 // Literals
-data class IntLiteral(val ctx: CellmataParser.Integer_literalContext, var value: Int = -1) : Expr()
+/**
+ * Represents a concrete integer value in the abstract syntax tree
+ */
+data class IntLiteral(val ctx: CellmataParser.Integer_literalContext, var value: Int = -1) : Expr(type = IntegerType)
```
falkecarlsen, line 163: I assume that `var value: Int = -1` defines the default value. If so, why choose -1, instead of 0?

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -74,39 +122,88 @@ data class NegativeExpr(val ctx: CellmataParser.NegativeExprContext, val value:
...
+ */
 data class IntToFloatConversion(val expr: Expr) : Expr()
 
+/**
+ * Internal node used to represent an implicit conversion from a array of states to a neighbourhood.
+ */
 data class StateArrayToActualNeighbourhoodConversion(val expr: Expr) : Expr()
 
 /*
  * Declarations
```
falkecarlsen, line 187: This should be an ordinary comment `//`, as I assume it refers to the category of the following definitions.

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -116,49 +213,114 @@ data class StateDecl(
...
 
 data class ForStmt(val ctx: CellmataParser.For_stmtContext, val initPart: AssignStmt, val condition: Expr, val postIterationPart: AssignStmt) : Stmt()
 
 data class BreakStmt(val ctx: CellmataParser.Break_stmtContext) : Stmt()
 
 data class ContinueStmt(val ctx: CellmataParser.Continue_stmtContext) : Stmt()
 
+/**
+ * Represents a become statement.
+ * The become statements terminates a state block, and changes the the state of the cell being evaluated.
```
falkecarlsen:
```suggestion
 * The become statements terminates a state block, and changes the state of the cell being evaluated.
```

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -4,6 +4,24 @@ import dk.aau.cs.d409f19.antlr.CellmataParser
 import org.antlr.v4.runtime.tree.ParseTree
 import org.antlr.v4.runtime.tree.TerminalNode
 
+/*
+ * Parse Tree to Abstract Syntax Tree transformer/mapper
+ *
+ * This file implements a recursive traversal of the the parse tree,
+ * where each step of the traversal emit a node in the AST if that node
```
falkecarlsen:
```suggestion
 * where each step of the traversal emits a node in the AST if that node
```

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -4,6 +4,24 @@ import dk.aau.cs.d409f19.antlr.CellmataParser
...
+ * in the parse tree is relevant to preserve.
+ * It's important to note that this mapper only handles the process of
+ * creating the tree part of the AST, it doesn't extract the values from
```
falkecarlsen, line 14: Well, it doesn't create the "tree-part", it *trims* the CST.

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -204,12 +240,22 @@ private fun reduceDecl(node: ParseTree): Decl {
...
+ * @throws AssertionError thrown when encountering an unexpected node in the parse tree.
+ */
 fun reduceCodeBlock(block: CellmataParser.Code_blockContext): List<Stmt> {
     return block.children
         .filter { it !is TerminalNode } // Remove terminals
         .map { reduceStmt(it) }
 }
 
+/**
+ * Attempts to recursively transform an ANTLR parse tree to an abstract syntax tree.
```
falkecarlsen: "Attempts to recursively transform an ANTLR parse tree to an abstract syntax tree." has been repeated thrice now, without mentioning that it reduces first a `stmt`, then a `decl`, lastly here the start-symbol.

I suggest that it is removed or elaborated.

compiler/src/main/kotlin/ast/symbol.kt
```diff
@@ -2,13 +2,26 @@ package dk.aau.cs.d409f19.cellumata.ast
 
 import java.util.*
 
+/**
+ * A single table in the symbol table.
+ * Table implements a nested design for a symbol table.
```
falkecarlsen, line 7:
```suggestion
 * Symbol table implements a nested design for a symbol table.
```

compiler/src/main/kotlin/ast/symbol.kt
```diff
@@ -79,4 +105,153 @@ class SymbolTable {
     fun openScope(index: Int) {
         scopeStack.push(scopeStack.peek().tables[index])
     }
+}
+
+/**
+ * Assists in building and filling a symbol table
```
falkecarlsen, line 111: Does not describe how it works at all. Maybe something like the following, but more elaborate:

It uses a stack, where tables are pushed and popped from, searches all tables on stack for getting, etc.

compiler/src/main/kotlin/main.kt
```diff
@@ -1,19 +1,24 @@
...
+    // Read from stress.cell
     val inputStream = ANTLRFileStream("src/main/resources/stress.cell")
+    // Setup lexer
     val lexer = CellmataLexer(inputStream)
     val tokenStream = CommonTokenStream(lexer)
+    // Setup parser
     val parser = CellmataParser(tokenStream)
 
+    // Run the lexer and parser
     val startContext = parser.start()
```
falkecarlsen, line 23:
```suggestion

// Trims the CST from the lexer/parser pass to an AST
```

compiler/src/main/kotlin/walkers/symbol.kt
```diff
@@ -3,14 +3,29 @@ package dk.aau.cs.d409f19.cellumata.walkers
...
 import java.lang.Exception
 
+/**
+ * Thrown when a undefined symbol is encountered. This exception indicates there is a use-before-declaration scenario.
+ */
 class SymbolException(val ident: String) : Exception("\"$ident\" was used before it was declared")
 
-class ScopeCheckVisitor(val symbolTable: SymbolTable) : BaseASTVisitor() {
+/**
+ * Walks through the abstract syntax tree, extracts symbols, and checks for use-before-declaration.
```
falkecarlsen, line 12: Missing explanation of what is meant by `symbolTableSession`, namely, what is meant by "session"?

compiler/src/main/kotlin/walkers/type.kt
```diff
@@ -0,0 +1,333 @@
...
+import dk.aau.cs.d409f19.cellumata.ast.*
+import kotlin.AssertionError
+
+/**
+ * Thrown when a violation of the type rules is found
+ */
+class TypeError : Exception()
+
+/**
+ * Move types up the abstract syntax tree according to the type rules, and check that there is no violation of the type rules
```
falkecarlsen: Rather "synthesizes", as this is what our SPO-course teaches as terminology?

compiler/src/main/kotlin/walkers/type.kt
```diff
@@ -0,0 +1,333 @@
...
+        node.setType(when {
+            node.left.getType() == BooleanType && node.right.getType() == BooleanType -> BooleanType
+            node.left.getType() == IntegerType && node.right.getType() == IntegerType -> BooleanType
+            node.left.getType() == FloatType && node.right.getType() == FloatType -> BooleanType
+            node.left.getType() == IntegerType && node.right.getType() == FloatType -> BooleanType
+            node.left.getType() == FloatType && node.right.getType() == IntegerType -> BooleanType
+            node.left.getType() == StateType && node.right.getType() == StateType -> BooleanType
+            node.left.getType() == ActualNeighbourhoodType && node.right.getType() == ActualNeighbourhoodType -> BooleanType
+            node.left.getType() == ActualNeighbourhoodType && node.right.getType() is ArrayType && (node.right.getType() as ArrayType).subtype == StateType -> BooleanType
+            node.left.getType() is ArrayType && (node.left.getType() as ArrayType).subtype == StateType && node.right.getType() == ActualNeighbourhoodType -> BooleanType
```
falkecarlsen, line 53: Hella long lines here and below, linebreak possible?

compiler/src/main/kotlin/walkers/value.kt
```diff
@@ -61,7 +78,9 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
 
         node.ident = node.ctx.neighbourhood_ident().text
 
+        // For each declared coordinate produce a Coordinate through the map operation
         val coords =  node.ctx.neighbourhood_code().coords_decl().map {
```
falkecarlsen: I think you forgot to reformat before shipping. I know the default rules mess with line comments following a statement, and this may be the reason as to why you didn't.
```suggestion
        val coords = node.ctx.neighbourhood_code().coords_decl().map {
```

compiler/src/main/kotlin/walkers/visitor.kt
```diff
@@ -341,11 +348,38 @@ abstract class BaseASTVisitor: ASTVisitor {
...
         // no-op
     }
 
     override fun visit(node: ReturnStmt) {
         visit(node.value)
     }
+}
+
+/**
+ * Walks the tree while opening and closing scopes has their are entered and left.
```
falkecarlsen: Their what? Something is missing.

falkecarlsen (changes requested):
Looks very good. I've put some pedantic comments regarding language, some regarding bad documentation, and other comments on possible linebreaks and choices for defaults.

compiler/src/main/kotlin/ast/ast.kt
jenrik: I choice `-1` to because it's different from the default value of a integer in kotlin, and to help aid in spotting errors while debugging. `-1` is intended to serve as indicator to the programmer that something _might_ be up.

compiler/src/main/kotlin/ast/ast.kt
jenrik: No, the "comment" is the documentation for `StateArrayToActualNeighbourhoodConversion`

compiler/src/main/kotlin/ast/symbol.kt
jenrik: "Table" refers to the class name

compiler/src/main/kotlin/ast/ast.kt
NicEastvillage: We could use `Int? = null`, which is a great way to indicate, that the value is not yet parsed. What do you think?

NicEastvillage (dismissed):
I see there's a few additions to the symbol table as well. I will let them through.

compiler/src/main/kotlin/ast/reduce.kt
jenrik: > Attempts to recursively transform an declaration node in the parse tree to create a subtree of the AST.

Would this be an acceptable phrasing for stmt, decl, and expr?

> Attempts to recursively transform the parse tree root node to an abstract syntax tree.

And how is this for the main `reduce` function?

compiler/src/main/kotlin/ast/symbol.kt
jenrik: The documentation could maybe be expanded, but I would like keep implementation details out of the javadoc since users of the class are not supposed to rely on the internal implementation.
I'm open to adding a comment inside the class that describe how it's supposed to work and how it's implemented.

compiler/src/main/kotlin/ast/reduce.kt
falkecarlsen: Yes, but use the stmt, decl, and expr where appropriate, e.g.;

For expression:
> Attempts to recursively transform an expression node in the parse tree to create a subtree of the AST.

I think you made a minor mistake with copy/paste in your reply, too.

compiler/src/main/kotlin/walkers/type.kt
jenrik:
```suggestion
 * Synthesizes types by moving them up the abstract syntax tree according to the type rules, and check that there is no violation of the type rules
```

compiler/src/main/kotlin/walkers/type.kt
jenrik: Suggestions are welcome

compiler/src/main/kotlin/walkers/visitor.kt
jenrik:
```suggestion
 * Walks the tree while opening and closing scopes as they are entered and left.
```

compiler/src/main/kotlin/walkers/symbol.kt
jenrik: > symbolTableSession is the symbol insertions session that keeps track of where to create scopes and insert symbols

How about this?

compiler/src/main/kotlin/ast/ast.kt
jenrik: I dislike the kotlin norm of using nullable as optional, but I guess I have to accept that it is the correct way to do it in kotlin

compiler/src/main/kotlin/ast/reduce.kt
jenrik:
```suggestion
 * Attempts to recursively transform the parse tree root node to an abstract syntax tree.
```
@cogitantium

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -182,9 +210,17 @@ private fun reduceStmt(node: ParseTree): Stmt {
     }
 }
 
+/**
+ * Attempts to recursively transform an ANTLR parse tree to an abstract syntax tree.
```
jenrik:
```suggestion
 * Attempts to recursively transform an declaration node in the parse tree to create a subtree of the AST.
```

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -135,10 +155,18 @@ private fun reduceExpr(node: ParseTree): Expr {
...
```
jenrik:
```suggestion
 * Attempts to recursively transform an statement node in the parse tree to create a subtree of the AST.
```

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -4,6 +4,24 @@ import dk.aau.cs.d409f19.antlr.CellmataParser
...
+ * in the parse tree is relevant to preserve.
+ * It's important to note that this mapper only handles the process of
+ * creating the tree part of the AST, it doesn't extract the values from
+ * the parse tree nodes, that is handled by ParseTreeValueWalker.
+ * This file handles the structure of the AST, and ParseTreeValueWalker
+ * handles the value of each node in the AST.
+ */
+
+/**
+ * Attempts to recursively transform an ANTLR parse tree to an abstract syntax tree.
```
jenrik:
```suggestion
 * Attempts to recursively transform an expression node in the parse tree to create a subtree of the AST.
```

compiler/src/main/kotlin/walkers/symbol.kt
falkecarlsen: On second thought, the existing is fine.

compiler/src/main/kotlin/walkers/type.kt
falkecarlsen: We may want to think about a distributed codestyle for IntelliJ, but not now. It's fine as is.

compiler/src/main/kotlin/main.kt
falkecarlsen: Actually, no.

compiler/src/main/kotlin/ast/ast.kt
jenrik: Now that I think about it we should fix this not only for `IntLiteral` but also for `BoolLiteral`, `FloatLiteral`, and `WorldDimension`. That fix should probably be a separate pull request since it's unrelated to documentation.

falkecarlsen: Agreed. Please create an issue for this.

jenrik: I have create issue #60

falkecarlsen: Nice. When @NicEastvillage approves, resolve discussion and merge PR.

falkecarlsen (approved):

NicEastvillage (approved):

PR #58, opened 2019-03-27
Remove positive expression from language

### LINKED ISSUE #48: Remove positive expression
The positive expression, exemplified by `+42` , has no effect, and is not part of standard mathematical notation. I suggest that we remove it.

falkecarlsen:
Removed positive expression from grammar, compiler, and `stress.cell`.
Resolves #48.


NicEastvillage (approved):
Looks good!

PR #59, opened 2019-03-27
Removed pre- and post increment/decrement

### LINKED ISSUE #52: Remove pre/post- increement/decrement operators
We decided to remove increment and decrement operators as they were expressions with side-effects, which becomes a bit complicated to make.

We can implement them later.

MathiasMehl:
Removed pre- and post increment/decrement from both grammar and compiler files. 
Removed all increments/decrements in stress.cell, and renamed variables
Resolves #52


NicEastvillage (changes requested):
I see that the grammar rule `modifiable_ident` is no longer used. Let's remove that too. Otherwise good job!

NicEastvillage (dismissed):

falkecarlsen (dismissed):

falkecarlsen (approved):

PR #61, opened 2019-03-28
Renamed some files so we don't have files with the same name.

NicEastvillage:
Renamed:
* package `walkers` to `visitors`
* `ast/symbol.kt` to `ast/symboltable.kt`
* `walkers/symbol.kt` to `visitors/symbolchecker.kt`
* `walkers/type.kt` to `visitors/typechecker.kt`
* `walkers/value.kt` to `visitors/valueextractor.kt`

We had two files named `type.kt` and two named `symbol.kt`. Now we don't, which means life should be less confusing.


falkecarlsen (approved):

jenrik:
I find it a little odd that that we have a class called `ScopeCheckVisitor` in a file called `symbolchecker.kt`, maybe we should rename either the class or the file to match the other.

PR #62, opened 2019-03-29
Fixed missing visit functions in ASTVisitor

jenrik:
Added new visit functions for FunctionArgs, Coordinate, WorldNode,
WorldDimension, ForStmt, BreakStmt, and ContinueStmt to ASTVisitor.
Added missing visit funtion to BaseASTVisitor for WorldNode,
WorldDimension, FunctionArgs, FloatLiteral, Coordinate, ForStmt,
BreakStmt, and ContinueStmt.
Fixed BaseASTVisitor#visit(NeighbourhoodDecl) not visiting coordinates.
Added value extraction for WorldNode and Coordinate.
Added missing body of "for" statement to ForStmt.
Added missing reduce call for "for" statement body.

### DIFF
--- a/compiler/src/main/kotlin/visitors/valueextractor.kt
+++ b/compiler/src/main/kotlin/visitors/valueextractor.kt
@@ -1,5 +1,6 @@
 package dk.aau.cs.d409f19.cellumata.walkers
 
+import dk.aau.cs.d409f19.antlr.CellmataParser
 import dk.aau.cs.d409f19.cellumata.ast.*
 
 open class ParsingException(msg: String = ""): Exception(msg)
@@ -151,5 +152,46 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
     override fun visit(node: AssignStmt) {
         super.visit(node)
         node.ident = node.ctx.var_ident().text
+    }
+
+    private fun parseInt(text: String, node: AST): Int {
+        try {
+            return text.toInt()
+        } catch (e: NumberFormatException) {
+            throw IntegerParsingException(text, node)
+        }
+    }
+
+    override fun visit(node: WorldNode) {
+        fun parseDimension(dim: CellmataParser.World_size_dimContext): WorldDimension {
+            val type = dim.type
+            return WorldDimension(
+                size = parseInt(dim.size.text, node),
+                type = when(type) {
+                    is CellmataParser.DimFiniteEdgeContext -> WorldType.EDGE // ToDo parse edge state
+                    is CellmataParser.DimFiniteWrappingContext -> WorldType.WRAPPING
+                    else -> throw AssertionError()
+                },
+                edge = if(type is CellmataParser.DimFiniteEdgeContext) {
+                    type.state.text
+                } else {
+                    null
+                }
+            )
+        }
+
+        val width = parseDimension(node.ctx.size.width)
+
+        node.dimensions = if (node.ctx.size.height != null) {
+            val height = parseDimension(node.ctx.size.height)
+            listOf(width, height)
+        } else {
+            listOf(width)
+        }
+
+        node.cellSize = parseInt(node.ctx.cellsize.value.text, node)
+        node.tickrate = parseInt(node.ctx.tickrate.value.text, node)
+
+        super.visit(node)
     }
 }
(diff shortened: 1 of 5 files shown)


compiler/src/main/kotlin/walkers/visitor.kt
```diff
@@ -337,24 +370,43 @@ abstract class BaseASTVisitor: ASTVisitor {
     }
 
     override fun visit(node: PreIncStmt) {
-        // no-op
+        visit(node.variable)
```
falkecarlsen: Remember that all pre/post, inc/dec statements have been removed. These are not present in neither the grammar nor compiler source on master.

compiler/src/main/kotlin/walkers/value.kt
```diff
@@ -152,4 +153,45 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
...
+
+    private fun parseInt(text: String, node: AST): Int {
+        try {
+            return text.toInt()
+        } catch (e: NumberFormatException) {
+            throw IntegerParsingException(text, node)
+        }
+    }
+
+    override fun visit(node: WorldNode) {
```
falkecarlsen: Please add some meaningful documentation to the following function.

falkecarlsen (dismissed):
Looks good, but there's an issue with your feature-branch having pre/post, inc/dec statements still implemented although master does not have these.

compiler/src/main/kotlin/visitors/valueextractor.kt
```diff
@@ -152,4 +153,54 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
...
+                    type.state.text
+                } else {
+                    null
+                }
+            )
+        }
+
+        val width = parseDimension(node.ctx.size.width)
+
+        // If height is non-null their is two dimension, otherwise their is only one
```
falkecarlsen:
```suggestion
        // If height is non-null it is two dimensional, otherwise it is one dimensional
```

compiler/src/main/kotlin/visitors/valueextractor.kt
```diff
@@ -152,4 +153,54 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
...
+         * Parses a single dimension from the world size declaration list
+         */
+        fun parseDimension(dim: CellmataParser.World_size_dimContext): WorldDimension {
+            val type = dim.type
+            return WorldDimension(
+                size = parseInt(dim.size.text, node),
+                type = when(type) {
+                    is CellmataParser.DimFiniteEdgeContext -> WorldType.EDGE
+                    is CellmataParser.DimFiniteWrappingContext -> WorldType.WRAPPING
+                    else -> throw AssertionError()
```
falkecarlsen, line 181: Should this not utilise our newly created ErrorLogger?

compiler/src/main/kotlin/visitors/valueextractor.kt
```diff
@@ -152,4 +153,54 @@ class LiteralExtractorVisitor : BaseASTVisitor() {
         super.visit(node)
         node.ident = node.ctx.var_ident().text
     }
+
+    /**
+     * Parse a string into a integer, or throw IntegerParisingException the string isn't a valid integer.
```
NicEastvillage:
```suggestion
     * Parse a string into a integer, or throw IntegerParsingException the string isn't a valid integer.
```

falkecarlsen (changes requested):
One question and one spelling-suggestion

compiler/src/main/kotlin/visitors/valueextractor.kt
NicEastvillage:
```suggestion
        // If height is non-null there are two dimensions, otherwise there is only one
```

compiler/src/main/kotlin/visitors/valueextractor.kt
jenrik: I don't think so, this is a case where if we hit the else statements it is not the user's fault, it's because the compiler is broken.

falkecarlsen: Okay, cool

falkecarlsen (approved):

NicEastvillage (approved):

PR #63, opened 2019-03-29
Graph printer

jenrik:
The graph printer will produce a graphviz graph of the abstract syntax
tree.
This PR depends on changes in #62


compiler/src/main/kotlin/walkers/grapher.kt
```diff
@@ -0,0 +1,359 @@
...
+    private fun printLabel(node: Any, label: String) {
+        output.println("\t" + node.hashCode() + " [label=\"" + label + "\"]")
+    }
+
+    override fun visit(node: WorldNode) {
+        var builder = StringBuilder()
+            .append("WorldNode\\nsize = ")
+            .append(node.dimensions.map {
+                when(it.type) {
+                    WorldType.EDGE -> it.size.toString() + "[edge=" + "]"
```
NicEastvillage: I think you are missing a the state name in `"[edge=" + "]"`

NicEastvillage (commented):
How do I test this?

jenrik:
I'm not sure what you mean by testing it, there isn't really a good way to automatically test this. If you wanna build a graphic from the output you can run the command `dot -Tsvg -o cellmata.svg ast.gs`, and it will produce a svg file

compiler/src/main/kotlin/walkers/grapher.kt
jenrik: fixed in 398bea9

compiler/src/main/kotlin/main.kt
```diff
@@ -40,6 +42,7 @@ fun compile(path: Path) {
         TypeChecker(symbolTable).visit(ast)
         ErrorLogger.assertNoErrors()
 
+        File("ast.gs").outputStream().use { out -> ASTGrapher(out).visit(ast) }
```
falkecarlsen, line 46: This should be an argument. Like #81, we should use a proper way of parsing arguments.

jenrik:
Here is a online graphviz render https://dreampuf.github.io/GraphvizOnline/

compiler/src/main/kotlin/visitors/grapher.kt
```diff
@@ -0,0 +1,305 @@
...
+
+        if (node.tickrate != null) {
+            builder = builder.append("\\ntickrate=").append(node.tickrate!!)
+        }
+
+        printLabel(node, builder.toString())
+        super.visit(node)
+    }
+
+    override fun visit(node: RootNode) {
```
falkecarlsen, line 53: Could we at least have the RootNode fun override be on top? I don't ask for a completely ordered hierarchy, but some structure could be nice.

compiler/src/main/kotlin/visitors/valueextractor.kt
```diff
@@ -2,6 +2,7 @@ package dk.aau.cs.d409f19.cellumata.visitors
 
 import dk.aau.cs.d409f19.antlr.CellmataParser
 import dk.aau.cs.d409f19.cellumata.ast.*
+import org.antlr.runtime.tree.ParseTree
```
falkecarlsen: Wait. This doesn't seem right? Why suddenly this import?

falkecarlsen (changes requested):
Minor errors and suggestions. Looks good though, but does not have tons of documentation - this is not strictly necessary.

compiler/src/main/kotlin/main.kt
jenrik: Let's move that change to a separate PR for commandline arguments

compiler/src/main/kotlin/visitors/grapher.kt
jenrik: This seem unimportant

compiler/src/main/kotlin/main.kt
falkecarlsen: Yes, that's fine. Maybe we should open an issue on it.

compiler/src/main/kotlin/visitors/grapher.kt
falkecarlsen: Well yes, but actually no. It's important for readability.

compiler/src/main/kotlin/main.kt
jenrik: I have create a issue, see #86

falkecarlsen (dismissed):

falkecarlsen (approved):

PR #65, opened 2019-04-04
Fixed null-list of children when body of program is empty

### LINKED ISSUE #66: Compiling a program with no body throws an NPE

falkecarlsen:
Trying to transform an empty list of children during reducing throws an `java.lang.IllegalStateException: node.body().children must not be null`. 

Fixed by checking if children count is zero, and returning an empty list instead of null.

Resolves #66.


falkecarlsen:
Note that error-reporting due to having a program with no body should happen during sanity checker, and not during reduce.

jenrik (commented):
You code is correct but, I would suggest writing this instead
`body = node.body().children?.map(::reduceDecl) ?: listOf()`

Maybe add a comment about how ANTLR produces null instead of empty lists

jenrik (approved):

PR #67, opened 2019-04-08
Removed builtin functions from symboltable reserved words

falkecarlsen:
This is a relic from when Cellmata should have builtin functions defined as reserved words.
Now that these should be user-defined functions, if necessary, this is faulty.

### DIFF
--- a/compiler/src/main/kotlin/ast/symboltable.kt
+++ b/compiler/src/main/kotlin/ast/symboltable.kt
@@ -25,7 +25,7 @@ class SymbolRedefinitionError(ctx: ParserRuleContext, val ident: String) : Error
 /**
  * List of language keywords that can't be used as identifiers
  */
-private val RESERVED_SYMBOLS: List<String> = listOf(
+private val RESERVED_WORDS: List<String> = listOf(
     "world",
     "neighbourhood",
     "state",
@@ -56,7 +56,7 @@ class SymbolTable {
     fun insertSymbol(ident: String, node: AST) {
         val table = scopeStack.peek()
 
-        if (RESERVED_SYMBOLS.contains(ident) || table.symbols.containsKey(ident)) {
+        if (RESERVED_WORDS.contains(ident) || table.symbols.containsKey(ident)) {
             ErrorLogger.registerError(SymbolRedefinitionError(node.ctx, ident))
         } else {
             table.symbols[ident] = node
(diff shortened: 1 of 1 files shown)


NicEastvillage (dismissed):
I disagree, that the functions should be user-defined (maybe I misunderstood?), they are still built-in. But yes, they should not be reserved symbols.

jenrik (changes requested):
I believe I discussed this change with @NicEastvillage some time ago, and we came to the the conclusion that builtin functions should not be reserved but instead be prefilled into the symbol table in the global scope, or a parent scope of the global scope. Unless we are removing builtin functions, this PR should also prefill the builtin functions into the symbol table.

falkecarlsen:
Maybe I've misunderstood then.

We agree that the builtin functions should not be reserved words, great. *Maybe we should rename the list from `RESERVED_SYMBOLS` to `RESERVED_WORDS` as the javadoc even says "keywords", too.*

If prefilling symbol table with builtin-function names when reaching symbol-checker phase, should then register an error, maybe `ReservedWordDefinitionError`, whenever it processes a state identifier or variable identifier which is either a reserved word or a builtin function. Agreed?

NicEastvillage:
@jenrik is describing what we need to do. The list of reserved keywords should contain "state, neighbourhood, for, break" ect. since the user cannot make a variable or function with these names. But if the user tries to call a function called `count`, then the symbol table should say "yes, that is a function, and here is the signature and definition." To achieve this, we need to prefill the symbol table with the builtin function, probably special AST nodes that derive from function nodes and code gen into specific functions.

So @jenrik this PR doesn't require changes. It is correct. But it doesn't add the functionality of built-in functions.

falkecarlsen:
I think that we all agree and I think that this PR should be merged as-is. 
41e6f3a45255af0eebbebcd4a70ed36a8f624595 contains the rework of builtin functions, where the symbol table is prefilled with all builtin functions. This required major rework of the AST classes which may indicate that the current structure is too inflexible and needs changes.

The point is that these two things; the error of including builtin functions in reserved words in the symbolchecker and implementing builtin functions correctly are separate. If desired, an issue could be created for this.

Let me know what you think, @jenrik and @NicEastvillage, on changing `RESERVED_SYMBOLS` to `RESERVED_WORDS`, and that'll be the last discussion on this PR, in my opinion.

jenrik (dismissed):

jenrik (approved):

NicEastvillage (approved):

PR #70, opened 2019-04-10
Sanitychecker

ebje17:
The sanity checker checks and throws an error if a become is in a function, a return is in a state. It also checks that a continue or a break is not outside a loop.
 It counts the number of states and if there is 0 it throws an error and if there is 1 it gives a warning.

### DIFF
--- a/compiler/src/main/kotlin/visitors/sanitychecker.kt
+++ b/compiler/src/main/kotlin/visitors/sanitychecker.kt
@@ -1,0 +1,88 @@
+package dk.aau.cs.d409f19.cellumata.visitors
+
+
+import dk.aau.cs.d409f19.cellumata.ErrorFromContext
+import dk.aau.cs.d409f19.cellumata.ErrorLogger
+import dk.aau.cs.d409f19.cellumata.ast.*
+import dk.aau.cs.d409f19.cellumata.walkers.BaseASTVisitor
+import org.antlr.v4.runtime.ParserRuleContext
+
+/**
+ * The error thrown when a become, continue, break or return statement is in a place where it should not be
+ */
+class SanityError(ctx: ParserRuleContext, description: String) : ErrorFromContext(ctx, description)
+
+/**
+ * A class to check that a become, return, break and continue statement does not appear in a place it should not be
+ */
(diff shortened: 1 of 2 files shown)


compiler/src/main/kotlin/visitors/sanitychecker.kt
```diff
@@ -0,0 +1,87 @@
...
+    // if there is 1 state it gives a warning
+    override fun visit(node: RootNode) {
+        super.visit(node)
+
+        if (numberOfStates == 0)
+            ErrorLogger.registerError(SanityError(node.ctx, "A state is needed"))
+        else if (numberOfStates == 1)
+            System.err.println("Warning: There is only one state")
+
+        println(numberOfStates)
```
falkecarlsen: This debug statement should probably not be here.

compiler/src/main/kotlin/visitors/sanitychecker.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    var inALoop = false
+
+    override fun visit(node: ForStmt) {
+        inALoop = true
+        super.visit(node)
+        inALoop = false
+    }
+
+    // Throws an error if a break is not in a loop
```
falkecarlsen:
```suggestion
    // Throws an error if a break is found outside a loop
```

Maybe this wording is better?

compiler/src/main/kotlin/visitors/sanitychecker.kt
```diff
@@ -0,0 +1,87 @@
...
+    override fun visit(node: BreakStmt) {
+        super.visit(node)
+        if (!inALoop)
+            ErrorLogger.registerError(SanityError(node.ctx, "Break statements are only allowed in loops"))
+    }
+
+    // Throws an error if a continue is not in a loop
```
falkecarlsen:
```suggestion
    // Throws an error if a continue is found outside a loop
```

falkecarlsen (changes requested):
Looks good! Minor comments to documentation and one debug print-statement.

falkecarlsen (approved):
Nice. It seems that all requested changes was fixed.

PR #72, opened 2019-04-11
Check if AssignStmt is a declaration.

jenrik:
Added a new field to AssignStmt that indicates that the statement is a
declaration and not just an assignment.
LiteralExtractorVisitor now check if AssignStmt is a declaration.


falkecarlsen (approved):
Looks good, this might be useful later on.

PR #73, opened 2019-04-11
Fixed array lookup using string

jenrik:
It should be possible to do an array lookup on any expression that could
contains a array. Previous behaviour limits you to naming variable that
should be looked through.


falkecarlsen (approved):
Nice, looks good. However, why are you prefixing your feature-branches with `pr-`?

jenrik:
@cogitantium I prefix the branch name with PR because I created it just for this pull request, the feature had already been developed on a separate branch

PR #74, opened 2019-04-11
Added return type to ASTVisitor

jenrik:
Allowing ASTVisitor implementations to return a value can help in certain use cases such as code generation.


NicEastvillage (commented):
Great idea. I have been thinking about this as well. I do think the BaseASTVisitor should be generic too though.

jenrik:
It's not really possible to to make `BaseASTVisitor` generic since it's uncertain what default value the visit functions should return

NicEastvillage (dismissed):
Oh yeah... In that case, approved.

NicEastvillage (approved):

PR #75, opened 2019-04-11
Fixed FuncExpr indentifier extraction

### LINKED ISSUE #55: The identifier in FuncExpr is the entire text of the function call, e.g. "baz(123, true)"
The identifier should only the name of the function and not the arguments. Otherwise it will never be found in the symbol table.

jenrik:
This pull request resolves #55


falkecarlsen (approved):
Nice. #worksonmymachinetoo

PR #76, opened 2019-04-11
Added code block node

### LINKED ISSUE #46: Add a code block node to AST
A code block node could replace the current use of `List<Stmt>` with a more well documented class. Code blocks could also be useful for inlining functions during optimization.

jenrik:
Replace the use of `List<Stmt>` with the `CodeBlock` node.
Generalised the visiting of code blocks/statement lists.
Resolves #46


NicEastvillage (approved):
Nice.

PR #77, opened 2019-04-11
Added names to types for better error messages

falkecarlsen:
This makes it so that `toString()` uses the fields of the type-classes, instead of the FQN of the class.

(cherry picked from commit 101f9bff2de321a92ad385543f020cb1c4b96b00)


NicEastvillage (approved):
I like this change. Wouldn't it be funny if I didn't?

PR #78, opened 2019-04-11
Added logic for checking correct returntype in function decleration

### LINKED ISSUE #64: TypeChecker aren't verifying function return type
The type checker doesn't verify the value of a return statement matches the declared return type of the function

MathiasMehl:
Added field to keep track of expected returntype from FuncDecl
If expected returntype does not match the value from ReturnStmt, register error
Implicit conversion from int to float in ReturnStmt visitor

Resolves #64
(cherry picked from commit b85f915731512b732b408a8e737e706ce8307880)


NicEastvillage (approved):
I can see that you are changing type of the expression node (the child node) of the return statement. I don't think that is what we want. Instead we want to insert a implicit conversion node (changing the AST by adding a node)... but I can see @jenrik hasn't done this anywhere. So I guess we will do this another time, unless you wanna play with it now.

PR #81, opened 2019-04-13
Made jar executeable and program should now be given as argument

NicEastvillage:
Added a run task to gradle, so we can run the compiler from gradle.

Also made the jar task produce a fat and executeable jar.

The .cell file which should be compiled should now be given as an argument. I.e. the compiled program is not hardcoded anymore. So if you want to run the compiler to compile `stress.cell` you should now either do
**1) use java**
run `gradlew jar`
followed by `java -jar build/libs/cellumata-1.0-SNAPSHOT.jar src/main/resources/compiling-programs/skeleton.cell`

**2) use gradle**
run `gradlew run -PappArgs="['src/main/resources/compiling-programs/skeleton.cell']"`

**3) configure intellij's run**
Setup intellij to run `gradlew run` and use the argument `-PappArgs="['src/main/resources/compiling-programs/skeleton.cell']"`
billede

We currently don't have other options for arguments, but I made them easy to add.


compiler/build.gradle
```diff
@@ -10,6 +10,25 @@ repositories {
...
+}
+
+jar {
+    manifest {
+        attributes 'Main-Class': 'dk.aau.cs.d409f19.cellumata.MainKt'
+    }
+
+    // This line of code recursively collects and copies all of a project's files
+    // and adds them to the JAR itself resulting in a 'fat jar'
+    from { configurations.compile.collect { it.isDirectory() ? it : zipTree(it) } }
```
jenrik, line 30: Doesn't gradle have builtin functions to handle the collection of classes?

jenrik (changes requested):

compiler/build.gradle
NicEastvillage: This is what stack overflow told me to do.

jenrik (approved):

PR #83, opened 2019-04-15
Clean-up of AST and ValueExtractor removed

### LINKED ISSUE #60: Use optional/nullable for unintialized literal values
Most AST nodes that stores values uses magic values to symbolize that they are uninitialized. It was proposed in the comments of #57 that we make the values nullable and set them to null to indicate that they are uninitialized.

### LINKED ISSUE #71: ValueExtractor assumes ordered existence of cellsize and tickrate although optional
Passing the Cellmata program:
```kotlin
world {
    size = 100 [wrap], 200 [wrap]
}

state ident (0, 0, 0) {
    become ident;
}
```

Throws an NPE:
```java
java.lang.NullPointerException
	at dk.aau.cs.d409f19.cellumata.visitors.LiteralExtractorVisitor.visit(valueextractor.kt:201)
	at dk.aau.cs.d409f19.cellumata.visitors.BaseASTVisitor.visit(visitor.kt:125)
	at dk.aau.cs.d409f19.cellumata.visitors.BaseASTVisitor.visit(visitor.kt:103)
	at dk.aau.cs.d409f19.Utilities$Companion.compileProgram(utility.kt:29)
	at dk.aau.cs.d409f19.BatchTest.batchPass(batchtest.kt:16)
```

NicEastvillage:
* AST is now more clean
* SourceContext is our new context, so now we only store what we need
* The ValueExtractor is removed and its functionality has been moved to `reduce.kt`

Resolves #71, indirectly resolves #60


compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -47,191 +64,252 @@ enum class WorldType {
...
-    val declaredType: Type? = UncheckedType
+    val declaredType: Type
 ) : Expr(ctx)
 
-data class ParenExpr(override val ctx: CellmataParser.ParenExprContext, val expr: Expr) : Expr(ctx)
-
-data class NamedExpr(override val ctx: ParserRuleContext, var ident: String = MAGIC_UNDEFINED_STRING) : Expr(ctx)
+class Identifier(
+    ctx: SourceContext,
+    var spelling: String
```
jenrik, line 194: Why was ident renamed to spelling?

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -150,47 +151,45 @@ private fun reduceStmt(node: ParseTree): Stmt {
...
         )
-        is CellmataParser.For_stmtContext -> ForStmt(
-            ctx = node,
-            initPart = AssignStmt(node.for_init().assignment(), expr = reduceExpr(node.for_init().assignment().expr())),
+        is CellmataParser.For_stmtContext -> ForLoopStmt(
+            ctx = SourceContext(node),
+            initPart = reduceStmt(node.for_init()) as AssignStmt,
             condition = reduceExpr(node.for_condition().expr()),
-            postIterationPart = AssignStmt(node.for_post_iteration().assignment(), expr = reduceExpr(node.for_post_iteration().assignment().expr())),
+            postIterationPart = reduceStmt(node.for_post_iteration()) as AssignStmt,
```
jenrik, line 183: The cast to `AssignStmt` will cause a crash if `ErrorStmt` is returned

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -29,114 +32,112 @@ import org.antlr.v4.runtime.tree.TerminalNode
...
         )
-        is CellmataParser.ParenExprContext -> ParenExpr(
-            ctx = node,
-            expr = reduceExpr(node.expr())
-        ) // ToDo: Should we flatten this?
-        is CellmataParser.LiteralExprContext -> reduceExpr(node.value)
-        is CellmataParser.VarExprContext -> NamedExpr(
-            ctx = node,
-            ident = node.ident.text
+        is CellmataParser.ParenExprContext -> reduceExpr(node.expr())
```
jenrik, line 112: Are we sure it's safe to remove parenthesis?

jenrik (changes requested):

compiler/src/main/kotlin/ast/reduce.kt
NicEastvillage: Yes. Parentheses just *change the precedence* and should only affect the parser, i.e. the parser builds a slightly different AST. After parsing the AST describes the evaluation order so it shouldn't have any consequences.

compiler/src/main/kotlin/ast/ast.kt
NicEastvillage: Well, NamedExpr are actually just identifiers (names of variables, constants, states, neighbourhoods), so I thought it was weird the names had a name. It also adheres to what we learned in SPO.

compiler/src/main/kotlin/ast/reduce.kt
NicEastvillage: The statement can only be an `AssignStmt` due to the grammar rules. If the statement is an `ErrorStmt` the error happened somewhere else, and we should already have raised an error.

jenrik (approved):

compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -47,191 +64,252 @@ enum class WorldType {
...
-    val ident: Expr,
+class ArrayLookupExpr(
+    ctx: SourceContext,
+    val arr: Expr,
     val index: Expr
 ) : Expr(ctx)
 
 /**
- * @param ctx Node in the parse tree this node corresponds to.
+ * @param ctx the context from witch this node was created from
```
falkecarlsen, line 182:
```
 * @param ctx the context from which this node was created from
```

falkecarlsen (commented):
Caught one spelling mistake

PR #90, opened 2019-04-16
Added BinaryExprs and updated visitor pattern

NicEastvillage:
Added sealed BinaryExpr, BinaryArithmeticExpr, BinaryBooleanExpr, and NumericComparisonExpr.
The binary expressions will allow us to generalise many things in the visitors, e.g the type checking for binary arithmetic expressions are basically identical.


jenrik (approved):
I'm curious to see how other people are gonna use this in practice

PR #91, opened 2019-04-16
Added scopes to for loops

NicEastvillage:


falkecarlsen (approved):
Looks good.

jenrik (approved):
Seems good, I assume that this is consistent with the semantic rules

PR #94, opened 2019-04-17
Include filename in batch-testing of Cellmata programs

falkecarlsen:
Refactored batch-testing of programs to include filename in error-message for better test-running output.


NicEastvillage (approved):
Looks good.

PR #96, opened 2019-04-23
Array size extraction

jenrik:
Adds code to find, propagate, and consistency check sized arrays and array literals. Additionally stress.cell have received additions for testing this new functionality and to remove broken code that prevented it from compiling.

This functionality will be used during code generation to allocate space for arrays.

### DIFF
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -206,13 +206,20 @@ class ArrayLookupExpr(
 
 /**
  * @param ctx the context from witch this node was created from
- * @param values Elements of the array.
+ * @param body Elements of the array.
  * @param declaredType Type listed before the body/values of the array.
  */
-class ArrayBodyExpr(
+class SizedArrayExpr(
+    ctx: SourceContext,
+    val body: ArrayLiteralExpr?,
+    val declaredType: Type,
+    var declaredSize: List<Int?>
+) : Expr(ctx)
+
+class ArrayLiteralExpr(
     ctx: SourceContext,
     val values: List<Expr>,
-    val declaredType: Type
+    var size: Int = -1
 ) : Expr(ctx)
 
 class Identifier(
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -219,7 +219,7 @@ class SizedArrayExpr(
 class ArrayLiteralExpr(
     ctx: SourceContext,
     val values: List<Expr>,
-    var size: Int = -1
+    var size: Int? = null
 ) : Expr(ctx)
 
 class Identifier(
--- a/compiler/src/main/kotlin/ast/reduce.kt
+++ b/compiler/src/main/kotlin/ast/reduce.kt
@@ -24,6 +24,11 @@ import org.antlr.v4.runtime.tree.TerminalNode
 val DEFAULT_CELLSIZE = 8
 val DEFAULT_TICKRATE = 10
 
+/**
+ * Get the size declared as part of the sized arrays type declaration.
+ *
+ * @return Returns a list of sizes from the type declaration, and null for any dimension which has to be inferred. Example "[10][][30]bool{}" would return 10, null, 30
+ */
 private fun getArrayDeclaredSize(array_decl: CellmataParser.Array_declContext): MutableList<Int?> {
     val l: MutableList<Int?> = if (array_decl.array_prefix().index != null) {
         mutableListOf(array_decl.array_prefix().index.text.toInt())
--- a/compiler/src/main/kotlin/ast/reduce.kt
+++ b/compiler/src/main/kotlin/ast/reduce.kt
@@ -24,6 +24,21 @@ import org.antlr.v4.runtime.tree.TerminalNode
 val DEFAULT_CELLSIZE = 8
 val DEFAULT_TICKRATE = 10
 
+private fun getArrayDeclaredSize(array_decl: CellmataParser.Array_declContext): MutableList<Int?> {
+    val l: MutableList<Int?> = if (array_decl.array_prefix().index != null) {
+        mutableListOf(array_decl.array_prefix().index.text.toInt())
+    } else {
+        mutableListOf(null)
+    }
+
+    // Assumption: Array is the only structured type
+    if (array_decl.type_ident() is CellmataParser.TypeArrayContext) {
+        l.addAll(getArrayDeclaredSize((array_decl.type_ident() as CellmataParser.TypeArrayContext).array_decl()))
+    }
+
+    return l
+}
+
 /**
  * Attempts to recursively transform an expression node in the parse tree to create a subtree of the AST.
  *
@@ -120,11 +135,15 @@ private fun reduceExpr(node: ParseTree): Expr {
             ident = node.value.ident.text
         )
         is CellmataParser.StateIndexExprContext -> StateIndexExpr(ctx = SourceContext(node))
-        is CellmataParser.ArrayValueExprContext -> ArrayBodyExpr(
+        is CellmataParser.ArraySizedValueExprContext -> reduceExpr(node.value)
+        is CellmataParser.Array_value_sizedContext -> SizedArrayExpr(
             ctx = SourceContext(node),
-            values = node.array_value().array_body().expr().map(::reduceExpr),
-            declaredType = typeFromCtx(node.value.type_ident())
+            declaredType = ArrayType(typeFromCtx(node.array_decl().type_ident())),
+            body = if(node.array_value_literal() == null) { null } else { reduceExpr(node.array_value_literal()) as ArrayLiteralExpr },
+            declaredSize = getArrayDeclaredSize(node.array_decl())
         )
+        is CellmataParser.ArrayLiteralExprContext -> reduceExpr(node.value)
+        is CellmataParser.Array_value_literalContext -> ArrayLiteralExpr(SourceContext(node), node.expr().map(::reduceExpr))
         is CellmataParser.LiteralExprContext -> reduceExpr(node.value)
         is CellmataParser.BoolLiteralContext -> BoolLiteral(
             ctx = SourceContext(node),
--- a/compiler/src/main/kotlin/ast/types.kt
+++ b/compiler/src/main/kotlin/ast/types.kt
@@ -18,7 +18,11 @@ object FloatType : Type("float")
 object BooleanType : Type("bool")
 object StateType : Type("state")
 object LocalNeighbourhoodType : Type("neighbourhood") // An evaluated neighbourhood
-data class ArrayType(val subtype: Type) : Type("array<$subtype>")
+data class ArrayType(val subtype: Type?) : Type("array<$subtype>") {
+    override fun toString(): String {
+        return "array<$subtype>"
+    }
+}
 
 /**
  * Default value for types before they're checked by the type checker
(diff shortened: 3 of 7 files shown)


compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -206,13 +206,20 @@ class ArrayLookupExpr(
...
+    val body: ArrayLiteralExpr?,
+    val declaredType: Type,
+    var declaredSize: List<Int?>
+) : Expr(ctx)
+
+class ArrayLiteralExpr(
     ctx: SourceContext,
     val values: List<Expr>,
-    val declaredType: Type
+    var size: Int = -1
```
NicEastvillage: Doesn't need a default value.

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -24,6 +24,21 @@ import org.antlr.v4.runtime.tree.TerminalNode
 val DEFAULT_CELLSIZE = 8
 val DEFAULT_TICKRATE = 10
 
+private fun getArrayDeclaredSize(array_decl: CellmataParser.Array_declContext): MutableList<Int?> {
```
NicEastvillage, line 32: Needs a good explaining javadoc

compiler/src/main/kotlin/ast/reduce.kt
```diff
@@ -24,6 +24,21 @@ import org.antlr.v4.runtime.tree.TerminalNode
...
+    val l: MutableList<Int?> = if (array_decl.array_prefix().index != null) {
+        mutableListOf(array_decl.array_prefix().index.text.toInt())
+    } else {
+        mutableListOf(null)
+    }
+
+    // Assumption: Array is the only structured type
```
NicEastvillage, line 39: I don't see why this is relevant?

compiler/src/main/kotlin/ast/types.kt
```diff
@@ -18,7 +18,11 @@ object FloatType : Type("float")
 object BooleanType : Type("bool")
 object StateType : Type("state")
 object LocalNeighbourhoodType : Type("neighbourhood") // An evaluated neighbourhood
-data class ArrayType(val subtype: Type) : Type("array<$subtype>")
+data class ArrayType(val subtype: Type?) : Type("array<$subtype>") {
+    override fun toString(): String {
```
NicEastvillage, line 22: Why is `toString()` overriden when Type's toString returns the same string? Did my trick not work?

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -233,25 +233,60 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
         node.setType(node.arr.getType())
     }
 
-    override fun visit(node: ArrayBodyExpr) {
+    /**
+     * Finds the size of the largest subarray in each dimension.
+     * Note: Note if their is empty array literals in the tree, this function may return fewer dimensions than expected.
```
NicEastvillage:
```suggestion
     * Note: if there are empty array literals in the tree, this function may return fewer dimensions than expected.
```

Also, shouldn't we avoid this behaviour?

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -233,25 +233,60 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+     */
+    private fun searchSize(node: ArrayLiteralExpr, sizes: MutableList<Int> = mutableListOf(), depth: Int = 1): MutableList<Int> {
```
NicEastvillage, line 240: You could use a local function for the recursion. This would remove the MutableList parameter. This way is also fine

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -233,25 +233,60 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+        node.values.forEach {
+            if (it is ArrayLiteralExpr) {
+                sizes = searchSize(it, sizes, depth + 1)
+            }
+        }
+
+        return sizes
+    }
+
+    private fun pushDownSize(n: Expr, size: List<Int>) {
```
NicEastvillage, line 257: Needs a javadoc

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -261,25 +296,73 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
             valuesType = types.first() // Pick any one because we have already checked they are identical
         }
 
+        /**
+         * Compares two types, handles nested checking for arrays.
+         * Null is equal to any other type.
+         */
+        fun compareType(type1: Type?, type2: Type?): Boolean {
```
NicEastvillage, line 305: `compareType`'s name and description does not tell me anything about its return type. I have to look at the code to determine that.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -261,25 +296,73 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+            }
+
+            return type1 == type2
+        }
+
         // ToDo implicit conversion, declaredType == float, and valuesType == integer -> type = float
         node.setType(when {
-            node.declaredType != null && node.values.isNotEmpty() -> {
-                if (node.declaredType != valuesType) {
+            node.body.values.isNotEmpty() -> {
```
NicEastvillage: You can use an `if`-expression instead of a `when` now. That way we don't need an else-case.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -261,25 +296,73 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+        val sizes = searchSize(node.body)
+
+        // compare sizes
+        if (sizes.size > node.declaredSize.size) {
+            throw AssertionError("Type consistency check has failed, it should have already caught this")
+        }
+        val finalSizes = node.declaredSize.mapIndexed { i, size ->
+            if (size == null && i >= sizes.size) {
+                null
+            } else if (size == null || (i < sizes.size-1 && sizes[i] > size)) {
```
NicEastvillage: This line is very hard to read. "size, sizes. sizes[i]?"

NicEastvillage: Shouldn't this case be illegal anyway? Because the literals size was greater than the declared size?

NicEastvillage (changes requested):
Lot's of small things. There should probably be some unit tests for this, at least some batch tests. Make sure it can be merged with master too.

compiler/src/main/kotlin/ast/ast.kt
jenrik:
```suggestion
    var size: Int? = null
```
Since the actual size isn't guaranteed to be know prior to type checking, we need to give it some value when it's instantiated during the reduce phase

compiler/src/main/kotlin/ast/types.kt
jenrik: If `toString()` isn't overriden the array type is incorrectly printed as `ArrayType(subtype=ArrayType(subtype=bool))`

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik: In the case of `[10][20][30]bool{}` searchSize would return a and list of length 1. I don't see how we would avoid this without forbidding empty array literals

compiler/src/main/kotlin/ast/reduce.kt
jenrik: I case we are gonna add support for structs we need to beware the this code will probably break

compiler/src/main/kotlin/ast/reduce.kt
jenrik:
```suggestion
/**
 * Get the size declared as part of the sized arrays type declaration.
 *
 * @return Returns a list of sizes from the type declaration, and null for any dimension which has to be inferred. Example "[10][][30]bool{}" would return 10, null, 30
 */
private fun getArrayDeclaredSize(array_decl: CellmataParser.Array_declContext): MutableList<Int?> {
```

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik:
```suggestion
    /**
     * Get the size declared as part of the sized arrays type declaration.
     *
     * @return Returns a list of sizes from the type declaration, and null for any dimension which has to be inferred. Example "[10][][30]bool{}" would return 10, null, 30
     */
    private fun pushDownSize(n: Expr, size: List<Int>) {
```

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik: What if I added this line to bottom of the javadoc?
`@return Returns true if the type are the same, accounting for wildcard matching`

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik: Fixed in d3251d1

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: That would work!

compiler/src/main/kotlin/ast/reduce.kt
NicEastvillage: That is good!

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: That's the same javadoc as `getArrayDeclaredSize()` ??

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: Oh then I misunderstood. As long as `{}` returns [0] and `{{{}}}` returns [1, 1, 0] even then it's `[10][10][10][10]bool{{{}}}`

compiler/src/main/kotlin/ast/ast.kt
NicEastvillage: That's true. I prefer `Int?` then

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik:
```suggestion
    /**
     * Updates an ArrayLiteral subtree with sizes.
     * Fill the size field of ArrayLiteral according to it's place in the tree and the values from size.
     */
    private fun pushDownSize(n: Expr, size: List<Int>) {
```
Sorry I copied the wrong javadoc

compiler/src/main/kotlin/visitors/typechecker.kt
jenrik: I regards to the naming, I'm not sure what to rename it to

NicEastvillage: What about
`sizes ==> literalSizes`
`size ==> declaredSize`

NicEastvillage (dismissed):

NicEastvillage (approved):

PR #97, opened 2019-04-26
Refactor of compile functions and adjusted tests accordingly

### LINKED ISSUE #95: Tests does not assert for errors between phases.
The tests currently uses the following function to compile programs:
```
/**
 * Compiles a Cellmata program given as string parameter
 */
fun compileProgram(program: String): CompilerData {
    val source = CharStreams.fromString(program)
    val lexer = CellmataLexer(source)
    val tokenStream = CommonTokenStream(lexer)
    val parser = CellmataParser(tokenStream)

    // Build AST
    val startContext = parser.start()
    val ast = reduce(startContext)

    // Symbol table and scope
    val scopeChecker = ScopeCheckVisitor()
    scopeChecker.visit(ast)
    val symbolTable = scopeChecker.getSymbolTable()

    // Type checking

    return CompilerData(parser, ast, symbolTable)
}
```
[Link to referenced code](https://github.com/d409f19/cellmata/blob/master/compiler/src/test/kotlin/utility.kt#L12-L34)

This contains a problem. It doesn't use the `ErrorLogger.assertNoErrors()`. What if there's an error in the program, and a phase cannot be completed? Well, when the following phase might throw exceptions due to missing analysis from earlier phases. This is the case on the `implicit-conversions`-branch. The `stress.cell` uses variables, which are undeclared. The scope checker finds these, but the compilation is not stopped, so the type checker will hit a null-pointer-exception which should never happen, causing a test to fail.

Fix this by adding the `ErrorLogger.assertNoErrors()` checks to this code similar to those in main and collect the compiler data available. A try-catch block should catch assertions from the ErrorLogger. Even better would be to use the same compile function for both production and test.

falkecarlsen:
Now both test and production uses the same compile-functions - mostly - since there's a need for unsafe compilation for some internal testing. Major refactoring of *production*-compilation has happened as a result.

These unsafe compile-functions for testing does not assert for no errors between phases. Since this may introduce strange exceptions, multiple unsafe compile-functions are introduced to only complete the compilation process to some extent.

Resolves #95

### DIFF
--- a/compiler/src/test/kotlin/symboltest.kt
+++ b/compiler/src/test/kotlin/symboltest.kt
@@ -65,7 +65,7 @@ class SymbolTest {
 
     /**
      * Returns the list of reserved words for testing data. Note that this cannot be passed to test by:
-     * @ValueSource("strings = RESERVED_WORDS") as JUnit5 complains that the source is not compileSource-time static
+     * '@ValueSource("strings = RESERVED_WORDS")' as JUnit5 complains that the source is not compile-time static
      */
     fun parserReservedSymbolsData(): List<String> {
         return RESERVED_WORDS
--- a/compiler/src/test/kotlin/symboltest.kt
+++ b/compiler/src/test/kotlin/symboltest.kt
@@ -43,11 +43,13 @@ class SymbolTest {
     fun assignStmtLiteralPassTest(value: String) {
 
         val compilerData =
-            compileProgram(getWorldDeclString() + "\n\n" + getStateDeclString(body = "let x = $value;"))
+            compileTestProgram(getWorldDeclString() + "\n\n" + getStateDeclString(body = "let x = $value;"))
+
+        // Ensure that symbol table actually exists
+        assertNotNull(compilerData.symbolTable)
 
         // Get SymbolTable for first subscope, which is first StateDecl
-        val stateSymbolTable = compilerData.symbolTable.tables[0]
-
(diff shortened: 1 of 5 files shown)


compiler/src/test/kotlin/symboltest.kt
```diff
@@ -63,7 +65,7 @@ class SymbolTest {
 
     /**
      * Returns the list of reserved words for testing data. Note that this cannot be passed to test by:
-     * @ValueSource("strings = RESERVED_WORDS") as JUnit5 complains that the source is not compile-time static
+     * @ValueSource("strings = RESERVED_WORDS") as JUnit5 complains that the source is not compileSource-time static
```
NicEastvillage: @ValueSource is the first symbol in the javadoc, so the line dissappears, which is not what you want. Also "compile-time" should not have been renamed to "compileSource-time". Be careful with comments when refractoring names with the rename function.

NicEastvillage (changes requested):
Approved ... except you messed up a comment.

compiler/src/test/kotlin/symboltest.kt
falkecarlsen: Yes, the rename-function can be a bit greedy. I never noticed that the actual rendering of the javadoc ate the annotation - added single-quotes to it in 6a07091.

NicEastvillage (approved):

PR #99, opened 2019-04-27
Pretty printer 

### LINKED ISSUE #93: Pretty printer

falkecarlsen:
Implemented complete pretty printer. 
Reproduces the source program faithfully* and matches indentation correctly when only one level is used. Covers all AST-nodes and is therefore complete.

*) Since for-loops make use of assignment-statements in both the initialisation- and post-iteration-parts, there will be an indentation before, and an extra linebreak after each assignment-statement.

The pretty-printed output of the updated `large-program.cell`.
```c
world {
	size = 1000 [edge=alive1], 1000 [wrap]
	tickrate = 100
	cellsize = 5
}

const boolean = false;

const integer = 42;

const floating = 42.42;

const goodProgram = true;

neighbourhood foobar {
	(-1, 1), (0, 1), (1, 1), (-1, 0), (1, 0)
}

state alive1 (0, 0, 0) {
	let a = -foo(42);
	become alive2;
}

state alive2 (0, 0, 0) {
	let a = -foo(42);
	become alive3;
}

state alive3 (0, 0, 0) {
	let a = -foo(42);
	become alive4;
}

state alive4 (0, 0, 0) {
	let a = -foo(42);
	become alive5;
}

state alive5 (0, 0, 0) {
	let a = -foo(42);
	let x = 0;
	for (	 x = 5;
x >= a;	 x = x + 1;
) {
	if (x != 0) {
	break;
	} else {
	continue;
	}
}
	become alive6;
}

state alive6 (0, 0, 0) {
	let a = -foo(42);
	let b = bar(a, true);
	if (b != -16) {
		become alive5;
	}
	become alive1;
}

function foo(int x) int {
	return x + 42;
}

function bar(int x, bool y) int {
	if (y) {
		return x + 42;
	} elif (!y) {
		return x;
	} else {
		return x + 4;
	}
}
```

Resolves #93

### DIFF
--- a/compiler/src/main/kotlin/main.kt
+++ b/compiler/src/main/kotlin/main.kt
@@ -32,11 +32,6 @@ fun compile(path: Path) {
         val sanityChecker = SanityChecker()
         sanityChecker.visit(ast)
 
-        // Build pretty-printed program
-        val prettyPrinter = PrettyPrinter()
-        prettyPrinter.visit(ast)
-        prettyPrinter.print()
-
         // Symbol table and scope
         val scopeChecker = ScopeCheckVisitor()
         scopeChecker.visit(ast)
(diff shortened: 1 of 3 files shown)


NicEastvillage (approved):
~~The indentation is sh~~ It looks good. We can improve it later.

PR #100, opened 2019-04-28
Fix ordering of world declaration options and add two test-cases

falkecarlsen:
Now the world declaration can take their options in any order, as long as they're declared after world size.

If more options are introduced at a later date, a rule on the form `options : (opt1 | opt2 | opt3 | ... )*` could be introduced, but this adds overhead in reduce phase and therefore the proposed solution is chosen.

### DIFF
--- a/compiler/src/main/antlr/Cellmata.g4
+++ b/compiler/src/main/antlr/Cellmata.g4
@@ -7,12 +7,16 @@ const_decl : STMT_CONST const_ident ASSIGN expr END ;
 const_ident : IDENT ;
 
 // World
-world_dcl : STMT_WORLD BLOCK_START size=world_size tickrate=world_tickrate? cellsize=world_cellsize? BLOCK_END ;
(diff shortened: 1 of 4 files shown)


jenrik (commented):
Seem fine except that I don't understand why world size has to be first

falkecarlsen:
@jenrik, yes I see your point. I briefly looked into implementing that, but the complexity rises a ton, as defining a grammar rule for that would allow repetitions (if possible combinations are not enumerated in a rule), which requires even more edge-case handling in the reduce-phase. 
If you still think this is an important feature, then please make an issue for it - I think we should put a pin in it for now.

jenrik:
I'm not a fan of doing a half job, I suggest we don't merge this as is

NicEastvillage (approved):
This is only half-done. But I don't want us to spend time on this. We have other features/bugs to fix. I will approve it though.

PR #104, opened 2019-04-30
Extension to SanityChecker with # only in states

ebje17:
Made a function in SanityChecker to check if a '#' is found outside a state.
Made two tests one where the '#' is in a state and one where it is in a function.


falkecarlsen (changes requested):
Please adhere to the naming-scheme of the other programs in the testing-directories.

E.g. your `stateIndexExpr-test.cell` would just be named `state-index-expr.cell` and then increment with `-1`, `-2` etc. if needed. 

Otherwise, it looks very nice!

falkecarlsen (approved):

PR #106, opened 2019-04-30
Fix nullable init and post-parts of for-loop

falkecarlsen:
Similarly to C, Cellmata should support `for (; x <= 5;) { }`

`AssignStmt` are now nullable, but this necessitates safe-calls in all places that visit the for-loop AST node, e.g., grapher and symbolchecker. Let-statements only visit these nodes if they're not null, else they ignore them.


NicEastvillage (approved):
Nice

PR #110, opened 2019-05-01
Fix broken visitor after array size changes

jenrik:
ArrayBodyExpr was replace by SizedArrayExpr and ArrayLiteralExpr which
meant also visitor has to updated.


falkecarlsen (approved):
Nice. Especially the pretty-printer change :smirk:

PR #111, opened 2019-05-01
Added settings. Currently pretty printing and graphing.

NicEastvillage:
This PR adds a data class for settings. The settings are currently very simple and will be changed when we get a more advanced argument parser. **The important thing in this PR is the CompilerSettings data class**. By adding it now we avoid multiple conflicts in the future when multiple features depends on the settings.


falkecarlsen (approved):
Nice, but there are many small quirks that I'm itching to fix.

jenrik (approved):

PR #112, opened 2019-05-01
Improved Error handling: Added Warnings and verbosity

NicEastvillage:
This PR refractors the ErrorLogger and adds warnings. The ErrorLogger is now also able to print errors with different levels of verbosity:
* standard verbosity is simply "Error at (32: 5): Some error messages" for both errors and warnings
* verbose (arg: -v) will make errors print their exact location in source context
* very verbose (arg: -vv) will make both errors and warnings print their exact location in source context

If an error or warning doesn't have a source context (ctx = null) the "at (l:c)" part is not printed.

Depends on #111


falkecarlsen (approved):
Nice. Best feature is that test-suite doesn't spit out a thousand warnings anymore.

PR #115, opened 2019-05-02
 Added edge option to world decl, that must be present when a dimension is an edge

### LINKED ISSUE #113: World edge should only allow one cell state; add an edge option to world declaration
Consider the program
```
world {
    size = 10 [edge=Alive], 10 [edge=Dead]
    ...
}
...
```
Which state are cells that are out of the grid in both dimensions?? Exactly, so let's change it to this:
```
world {
    size = 10 [edge], 10 [edge]
    edge = Dead
    ...
}
...
```
allowing only one kind of edge state.

NicEastvillage:
Also adds semi-colons after each world option

Resolves #113


compiler/src/main/kotlin/visitors/prettyprinter.kt
```diff
@@ -58,7 +58,7 @@ class PrettyPrinter : BaseASTVisitor() {
                 "${node.dimensions[n].size} ${if (node.dimensions[n].type == WRAPPING) {
                     "[wrap]$separator"
                 } else if (node.dimensions[n].type == EDGE) { // Note the non-null assertion, required for field access
-                    "[edge=${node.dimensions[n].edge!!.spelling}]$separator"
+                    "[edge$separator"
```
falkecarlsen:
```suggestion
                    "[edge]$separator"
```

Missing a closing square bracket

falkecarlsen (changes requested):
Looks fine. Just one thing that seems wrong. Please do try an pretty print to see if it works well. Use `large-program.cell` maybe.

compiler/src/main/kotlin/visitors/prettyprinter.kt
NicEastvillage: Fixed in 958249b

NicEastvillage:
@cogitantium this is the pretty print result now. The for-loop still has an extra semicolon, and the indentation is weird, but otherwise everything is good.
```
world {
	size = 1000 [edge], 1000 [wrap];
	edge = alive1;
	tickrate = 100;
	cellsize = 5;
}

const boolean = false;

const integer = 42;

const floating = 42.42;

const goodProgram = true;

neighbourhood foobar {
	(-1, 1), (0, 1), (1, 1), (-1, 0), (1, 0)
}

state alive1 (0, 0, 0) {
	let a = -foo(42);
	become alive2;
}

state alive2 (0, 0, 0) {
	let a = -foo(42);
	become alive3;
}

state alive3 (0, 0, 0) {
	let a = -foo(42);
	become alive4;
}

state alive4 (0, 0, 0) {
	let a = -foo(42);
	become alive5;
}

state alive5 (0, 0, 0) {
	let a = -foo(42);
	let x = 0;
	for (	 x = 5;
x >= a;	 x = x + 1;
) {
	if (x != 0) {
		 x = -5;
break;
	} else {
		 x = 42 + 3 / 176 * 22;
continue;
	}
}
	become alive6;
}

state alive6 (0, 0, 0) {
	let a = -foo(42);
	let b = bar(a, true);
	if (b != -16) {
		become alive5;
	}
	become alive1;
}

function foo(int x) int {
	return x + 42;
}

function bar(int x, bool y) int {
	if (y) {
		return x + 42;
	} elif (!y) {
		return x;
	} else {
		return x + 4;
	}
}
```

falkecarlsen (approved):

PR #119, opened 2019-05-02
Implemented flowcheck

### LINKED ISSUE #80: Check that a function has a return in all branches
As of now, it is possible for a function to have a branch without a return, e.g 
```kotlin
function f3(int a, int b) bool {
    let c = true;
    for (let x = a; x < b; x = x+1) {
        if ( x == 5 ) {
            return c && false;
        }
    }
}
```
And this should produce an error

MathiasMehl:
flowchecker will produce warnings if code is blocked by either a break or return in function
If statements will have to have either return/break/continue in all branches to produce a warning. 

A FlowError will be produced if it is not possible for a function to meet a return

Resolves #80


compiler/src/main/kotlin/main.kt
```diff
@@ -101,6 +101,11 @@ fun compile(source: CharStream, settings: CompilerSettings): CompilerData {
     TypeChecker(symbolTable).visit(ast)
     ErrorLogger.assertNoErrors()
 
+    // Flow checking
```
NicEastvillage: Flow checking should be done after sanity check and before symbol check. That's what we wrote on the blackboard, and I think that makes more sense.

compiler/src/main/kotlin/visitors/flowchecker.kt
```diff
@@ -0,0 +1,445 @@
...
+        var containsBreak = false
+
+        //visit all Stmts in the body, index used to print line number
+        for (i in 0 until node.body.size) {
+
+            val flow = visit(node.body[i])
+
+            //codeAfterThisBlock asserts that there is at least 2 Stmts in the body. No warnings will be
+            //produced if the current Stmt is the last, as it can not possibly block any further code inside the block
+            val codeAfterThisBlock = i < node.body.size - 1 || i == 0 && 1 < node.body.size
```
NicEastvillage: I don't think the `i == 0 && 1 < node.body.size` is necessary. If `i == 0 && 1 < node.body.size` is true, then `i < node.body.size - 1` is true as well.

compiler/src/main/kotlin/visitors/flowchecker.kt
```diff
@@ -0,0 +1,445 @@
...
+            if (codeAfterThisBlock) when {
+                //if the visited node has a return that blocks further code in the block, produce warning
+                containsReturn -> {
+                    ErrorLogger += FlowWarning(
+                        node.body[i].ctx,
+                        "Code after statement never met, reason: return statement"
+                    )
+                    /*if a return blocking code is met, return. If this is the first Stmt in the codeblock
+                    the function still is guaranteed to meet a return, else it is not guaranteed to meet a return*/
+                    return if (i == 0) Flow(containsReturn = true) else Flow(containsReturn = false)
```
NicEastvillage: Why is there an if-statement here? The code finds out that a Stmt guarentees to meet a return, when it doesn't matter if the Stmt is the first or the second etc.

compiler/src/main/kotlin/visitors/flowchecker.kt
```diff
@@ -0,0 +1,445 @@
...
+        return visit(node.state)
+    }
+
+    //if a ReturnStmt is met, return true
+    override fun visit(node: ReturnStmt): Flow {
+        visit(node.value)
+        return Flow(containsReturn = true)
+    }
+
+    override fun visit(node: ForLoopStmt): Flow {
```
NicEastvillage, line 378: Why does this function not contain any logic? We can't guarantee that the body of a for-loop is executed, so it should always return `Flow(false, false)`.

compiler/src/main/kotlin/visitors/flowchecker.kt
```diff
@@ -0,0 +1,445 @@
...
+                    ErrorLogger += FlowWarning(
+                        node.body[i].ctx,
+                        "Code after statement never met, reason: return statement"
+                    )
+                    /*if a return blocking code is met, return. If this is the first Stmt in the codeblock
+                    the function still is guaranteed to meet a return, else it is not guaranteed to meet a return*/
+                    return if (i == 0) Flow(containsReturn = true) else Flow(containsReturn = false)
+                }
+                //if the visited node has a break/continue that blocks further code in the block, produce error
+                containsBreak && node.body[i] !is ForLoopStmt -> {
```
NicEastvillage, line 434: This case is wrong. I can write:
```
if (x) {
    break;
} else {
    break;
}
```
and the result will be `Flow(false, false)` even though all branches contains a break. I will try to find a program, where this produces the wrong error...

NicEastvillage (changes requested):
Most of it looks good, but you forgot one thing. We aren't guaranteed to enter the body of a for-loop. But currently the following function compiles:
```
function f1() int {

    for (let i = 0; false; i = i + 1) {
        return 0;
    }

    let reachable = true;
}
```
with warning: `Warning at (for-loop): Code after statement never met, reason: return statement`

compiler/src/main/kotlin/visitors/flowchecker.kt
NicEastvillage: Found a function with wrong warnings:
```
function f2() int {

    for (let i = 0; false; i = i + 1) {
        if (true) { // No warning, but should give one!
            if (true) { // Warning: Code after statement never met, reason: break/continue statement
                break;
            } else {
                break;
            }

            let b = false;
        } else {
            if (true) { // Warning: Code after statement never met, reason: break/continue statement
                break;
            } else {
                break;
            }

            let b = false;
        }

        let reachable = false;
    }

    return 0;
}
```

compiler/src/main/kotlin/visitors/flowchecker.kt
MathiasMehl: fixed in commit ba22229 
Now For-loops always returns (false,false)

compiler/src/main/kotlin/main.kt
MathiasMehl: flowcheck phase moved in commit cb58b71

compiler/src/main/kotlin/visitors/flowchecker.kt
MathiasMehl: expression simplified in commit 945f7cf

compiler/src/main/kotlin/visitors/flowchecker.kt
MathiasMehl: removed the if statement, and now always returns true in commit 026f208

compiler/src/main/kotlin/visitors/flowchecker.kt
MathiasMehl: the returnvalue in the case that the stmt contains a break, is now set to Flow(containsReturn, containsBreak), so now the warning will flow up and produce more warnings.
commit e5f57895

Meaning this code now produces 3 warnings

NicEastvillage (approved):

PR #120, opened 2019-05-03
Add negative floats-tests to AST-tests

### LINKED ISSUE #84: Negative literal floats are converted to their absolute value
Whenever an expression contains a negative float literal, it's converted to the absolute float value, e.g; `x = -42.2`. The value of this expression on the D-AST is `42.2`.

falkecarlsen:
Since negative float literals are now correctly represented, negative float literals has been added to AST-tests.

Resolves #84

### DIFF
--- a/compiler/src/test/kotlin/asttest.kt
+++ b/compiler/src/test/kotlin/asttest.kt
@@ -237,7 +237,6 @@ class ASTTest {
 
         /**
          * Test constant declaration float values. Assumes that default world declaration passes.
-         * TODO: Does not include negative values as they fail! When fixed, they should be implemented again.
          */
         @ParameterizedTest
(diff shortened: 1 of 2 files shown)


compiler/src/test/kotlin/asttest.kt
```diff
@@ -237,10 +237,9 @@ class ASTTest {
 
         /**
          * Test constant declaration float values. Assumes that default world declaration passes.
-         * TODO: Does not include negative values as they fail! When fixed, they should be implemented again.
          */
         @ParameterizedTest
-        @ValueSource(strings = ["3.14159", "1000.99", "123456.789"])
+        @ValueSource(strings = ["3.14159", "1000.99", "123456.789", "-42.0000123401234", "-42.00001234012340"])
```
NicEastvillage, line 242: What about "0.0"

NicEastvillage (approved):

PR #122, opened 2019-05-03
Fixed missing scope around else clause.

### LINKED ISSUE #121: Scope error in else-blocks
The following program does not compile
```
world {
    size = 100 [wrap], 100 [wrap]
    cellsize = 8
    tickrate = 3
}

state alive (0, 0, 0) {
    if (true) {
        let x = 0;
    } else {
        let x = 1;
    }
    let x = 2;
}
```
due to `Error at (13:4): "x" is already defined` (the x after the if-statement).

I believe it is caused by wrong scoping of the else-blocks.

NicEastvillage:
Resolves #121


falkecarlsen (approved):
Looks good

PR #124, opened 2019-05-03
Implement builtin functions

### LINKED ISSUE #49: Add builtin function to symbol table
The builtin functions should be added to the symbol table. Currently the builtin functions are part of a list of reserved words that are blocked from being inserted into the symbol table.

falkecarlsen:
- Builtin function nodes
- Symbol-checking for builtin functions and creating nice error-messages
- Examples which tests this functionality

Resolves #49


compiler/src/main/kotlin/ast/ast.kt
```diff
@@ -456,12 +559,13 @@ class ReturnStmt(
...
     ctx: SourceContext,
     val body: List<Stmt>
-): AST(ctx)
+) : AST(ctx)
 
 /*
  * Error nodes are used when something goes wrong in reduce.kt and is returned by the failing function.
  */
 class ErrorAST : AST(EMPTY_CONTEXT)
+
```
NicEastvillage: 😠 
```suggestion
```

compiler/src/main/kotlin/ast/symboltable.kt
```diff
@@ -109,6 +122,16 @@ class CreatingSymbolTableSession(symbolTable: Table) {
     init {
         scopeStack.push(Table(tables = mutableListOf(symbolTable))) // God scope
         scopeStack.push(symbolTable)
+        //TODO: preload symboltable with builtin functions
```
NicEastvillage:
```suggestion
```

NicEastvillage (changes requested):
The argument names are weird and inconsistent, but whatever.

I think you should add some compiling batch tests to check the arguments of built-in functions.

compiler/src/main/kotlin/ast/ast.kt
falkecarlsen: Turn that frown upside down with this 2302085

compiler/src/main/kotlin/ast/symboltable.kt
falkecarlsen: Refactored preloading and removed that todo in 36794db

falkecarlsen:
Function parameters has been refactored in 85cf411.

Both compiling and failing tests on parameter passing has been added.

Type-checker has been expanded with function-call argument type-checking against formal arguments of function declaration. This works well and appropriate tests have been added. Relevant commit 8091472. Note that a ton of minor reformatting has been done too.

#125 was found during the expansion of the type-checker. I'm not sure how to fix it, hence the issue.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -408,6 +457,34 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
     override fun visit(node: FuncCallExpr) {
         super.visit(node)
 
+        // Get function declaration for type-checking
+        val funcDecl = (symbolTableSession.getSymbol(node.ident) as FuncDecl)
```
NicEastvillage: The identifier might not be a FuncDecl so this could creating casting exceptions. This should also be checked by the typechecker, and it should raise an error saying something like "ident cannot be called since it is not a function". It would be relevant in this program:
```
// omitted
state Alive (0, 0, 0) {
    let x = 0;
    let y = x();
}
```

NicEastvillage (changes requested):
Nice. Didn't know that we didn't have type-checking code for arguments yet. Good job adding them. I can only spot one problem.

I think I know how to fix that issue #125. I can look into it later.

compiler/src/main/kotlin/visitors/typechecker.kt
falkecarlsen: Fixed in 9314955. Has test exactly as suggested.

falkecarlsen:
Cool. Let me know about #125, I pulled my hair out trying to figure that out.

Now prints `Error at (9:12): "x" cannot be called, as it is not a function` on calling a variable as a function.

NicEastvillage (approved):

PR #127, opened 2019-05-06
Implicit conversions

### LINKED ISSUE #89: Add ImplicitConversion nodes to AST when needed in type checker
Currently we either just allow integers where we want floats or throw an error, but ideally we should insert a conversion node to decorate the integer branch. Same goes for arrays of states and neighbourhoods.

### LINKED ISSUE #118: Typechecker int to bool
The following code does not produce a TypeError, even though the return type is wrong

function f(int a) bool {
    return a;
}

### LINKED ISSUE #125: Local neighbourhoods are not given a type
Program `builtin-count.cell`;
```
world {
    size = 100 [wrap], 200 [wrap]
    tickrate = 1
    cellsize = 5
}

neighbourhood pseudomoore {
    (-1, 1), (0, 1), (1, 1),
    (-1, 0),          (1, 0),
    (-1, -1), (0, -1), (1, -1)
}

state ident (0, 0, 0) {
    // Number of 'ident' neighbours in pseudomoore neighbourhood
    let x = count(ident, pseudomoore);
    become ident;
}
```

fails with;
```
Error at (15:12): Actual argument "pseudomoore", of type null, was not equal to formal argument "neighbourhoodParameter", of type neighbourhood
```
since  local neighbourhoods' types are never set.

NicEastvillage:
Adds implicit conversion. When a construct expects a certain type and the actual type does not match, it will now try to do implicit conversion.

Constructs with implicit conversion:
* NumericComparison
* BinaryArithmeticExpr
* Return type
* Function call arguments
* Array initialisation
* Assignment

Resolves #89, resolves #125, resolves #118

Implicit conversion contains one error due to an earlier mistake of mine. That is `==` only accepts floats and ints, but other types should also be comparable. See issue #131


compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -292,166 +165,243 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
-                when {
-                    node.value.getType() == IntegerType -> FloatType
-                    else -> {
-                        ErrorLogger += TypeError(
-                            node.ctx,
-                            "Wrong return type (${node.value.getType()}). Expected $expectedReturn"
-                        )
-                        null
-                    }
+        expectedReturn!!
```
falkecarlsen, line 399: Please write a comment as per discussion in group. It's very hidden as to why there's a null-check here.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -15,274 +14,148 @@ class TypeError(ctx: SourceContext, description: String) : CompileError(ctx, des
...
-                    null
-                }
-            }
-        )
-    }
-
-    override fun visit(node: GreaterThanExpr) {
-        super.visit(node)
+        if (node.value.getType() != BooleanType) {
+            ErrorLogger += TypeError(node.ctx, "Can only NOT boolean expressions.")
```
falkecarlsen: "Can only NOT boolean expressions." :thinking: Please revise.

falkecarlsen: I understand what it means, but maybe it could be written nicer.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -15,274 +14,148 @@ class TypeError(ctx: SourceContext, description: String) : CompileError(ctx, des
...
-            }
-        )
-    }
+    /**
+     * Finds the type of an array based on its content. No types are changed converted, but errors are logged if the
+     * array's type (or if a sub-array's type) can't be determined.
+     * @see pushDownArrayType
+     */
+    private fun searchArrayType(expr: Expr): Type {
+        if (expr !is ArrayLiteralExpr) {
```
falkecarlsen: Kotlin suggests a when-expression for the cascade if-statements.

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -15,274 +14,148 @@ class TypeError(ctx: SourceContext, description: String) : CompileError(ctx, des
...
-                node.left.getType() == FloatType && node.right.getType() == IntegerType -> BooleanType
-                else -> {
-                    ErrorLogger += TypeError(node.ctx, "Right and left hand side of must be either float or integer.")
-                    null
-                }
-            }
-        )
-    }
+    /**
+     * Finds the type of an array based on its content. No types are changed converted, but errors are logged if the
```
falkecarlsen: Choose either 'changed' or 'converted'. Not both

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -15,274 +14,148 @@ class TypeError(ctx: SourceContext, description: String) : CompileError(ctx, des
...
-                    null
-                }
-            }
-        )
-    }
+            expr.setType(type)
 
-    override fun visit(node: NegationExpr) {
-        super.visit(node)
+            if (type.subtype == FloatType) {
```
falkecarlsen: Kotlin suggests a when-expression for the cascade if-statements. Looks better in my opinion.

falkecarlsen (changes requested):
Looks good. Minor suggestions and I'd very much like for you to reformat the `typechecker.kt` with IDEA.

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: Good idea. See d783de7

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: Changed it to "The NOT-operator only works on boolean expressions." in d783de7

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: Jokes on you, because I meant both; "changed *or* converted"

falkecarlsen (approved):

PR #130, opened 2019-05-06
Add git information to compiler jar

jenrik:
These changes adds git status information, including last commit, current branch, whether the working tree is dirty, and the time of build, to the jar manifest. This makes the any distributed compiler version traceable, and there for easier to debug.


NicEastvillage (approved):
This does not seem super important for a school project, but sure, now we have it!

Issue #139, opened 2019-05-07
Cellsize can be negative


MathiasMehl:
No errors in the following program

world {
size = 100 [edge], 200 [wrap];
edge = ident;
cellsize = -4;
}

state ident (0, 0, 0) {
become ident;
}

Issue #138, opened 2019-05-07
Tickrate can be 0


MathiasMehl:
The tickrate can be set to 0, this should produce either a warning or an error, as the rendering would be quite boring.

world {
size = 100 [edge], 200 [wrap];
edge = ident;
tickrate = 0;
}

state ident (0, 0, 0) {
become ident;
}

Issue #137, opened 2019-05-07
Negative tickrate


MathiasMehl:
The following programs compiles with no warnings or errors regarding the tickrate which is negative. Which makes no sense.

world {
size = 100 [edge], 200 [wrap];
edge = ident;
tickrate = -1;
}

state ident (0, 0, 0) {
become ident;
}

Issue #136, opened 2019-05-07
Negative size in dimension


MathiasMehl:
The following programs compiles with no warnings or errors regarding the size

world {
    size = -1 [edge], 200 [wrap];
    edge = ident;
}

state ident (0, 0, 0) {
    become ident;
}

PR #142, opened 2019-05-07
World field sanity

MathiasMehl:
Added logic in sanitychecker.kt to add errors, if cellsize, tickrate, or dimension sizes are 0 or negative. 

solves issues #136 #137  #138  #139


compiler/src/main/kotlin/visitors/sanitychecker.kt
```diff
@@ -31,6 +31,29 @@ class SanityChecker : BaseASTVisitor() {
         if (worldHasEdge && node.edge == null) {
             ErrorLogger += SanityError(node.ctx, "A dimension is an [edge], but the edge state was not declared.")
         }
+
+        if (node.cellSize <= 0) {
+            ErrorLogger += SanityError(
+                node.ctx, "Cellsize must be an integer larger than 0, but was found to be" +
```
falkecarlsen:
```suggestion
                node.ctx, "Cellsize must be an integer larger than 0, but was: " +
```

Setting the value after a colon indicates that it is an internal value moreso than something written directly into error-template. 

This applies for the following two.

falkecarlsen (changes requested):
Minor thing, otherwise very nice.

falkecarlsen (approved):

PR #145, opened 2019-05-07
Improved batch test failing output

### LINKED ISSUE #143: BatchFailTest displays useless error message if the compiler crashes
If the compilation of a program results in the compiler crashing, the test fails as it should, but it is not possible to see which file was being compiled and why it crashed.

falkecarlsen:
Now prints a stack-trace if compiler fails spectacularly on batch-testing failing programs.

Resolves #143

### DIFF
--- a/compiler/src/test/kotlin/batchtest.kt
+++ b/compiler/src/test/kotlin/batchtest.kt
@@ -111,22 +111,26 @@ class BatchTest {
                     compileTestProgram(program)
                 }
             } catch (e: AssertionFailedError) {
-                // If compilation does not throw a TerminatedCompilationException,
-                // compile again and assert for errors in parser
-                val compileData = compileTestProgram(program)
+                try {
+                    // If compilation does not throw a TerminatedCompilationException,
+                    // compile again and assert for errors in parser
+                    val compileData = compileTestProgram(program)
 
-                // Assert that no errors are contained in ErrorLogger, as this would indicate a fault with asserting for no errors
-                assertFalse(
-                    ErrorLogger.hasErrors(),
-                    "Non-compiling program failed spectacularly! Errorlogger initially didn't throw an " +
-                            "TerminatedCompilationException, yet the ErrorLogger contains errors!"
-                )
+                    // Assert that no errors are contained in ErrorLogger, as this would indicate a fault with asserting for no errors
+                    assertFalse(
+                        ErrorLogger.hasErrors(),
+                        "Non-compiling program failed spectacularly! Errorlogger initially didn't throw an " +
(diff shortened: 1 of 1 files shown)


compiler/src/test/kotlin/batchtest.kt
```diff
@@ -111,22 +111,26 @@ class BatchTest {
...
+                        "Non-compiling program failed spectacularly! Errorlogger initially didn't throw an " +
+                                "TerminatedCompilationException, yet the ErrorLogger contains errors!"
+                    )
+                    // Assert that parsing errors are found
+                    assertTrue(
+                        compileData.parser.numberOfSyntaxErrors > 0,
+                        "Non-compiling program had neither any TerminatedCompilationException nor parsing-errors! Filename: $filename"
+                    )
+                } catch (e: Exception) {
+                    fail("Spectacular error in compiler. Printing stacktrace of: ${e.javaClass}", e)
```
jenrik, line 131: Are you sure this prints the stacktrace?
https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#toString--

compiler/src/test/kotlin/batchtest.kt
```diff
@@ -111,22 +111,26 @@ class BatchTest {
...
+                        ErrorLogger.hasErrors(),
+                        "Non-compiling program failed spectacularly! Errorlogger initially didn't throw an " +
+                                "TerminatedCompilationException, yet the ErrorLogger contains errors!"
+                    )
+                    // Assert that parsing errors are found
+                    assertTrue(
+                        compileData.parser.numberOfSyntaxErrors > 0,
+                        "Non-compiling program had neither any TerminatedCompilationException nor parsing-errors! Filename: $filename"
+                    )
+                } catch (e: Exception) {
```
jenrik, line 130: I'm not sure if it's safe to catch `Exception`, would this catch `OutOfMemoryError`?

falkecarlsen: https://docs.oracle.com/javase/7/docs/api/java/lang/OutOfMemoryError.html

That is not an exception, so I guess the point is moot.

compiler/src/test/kotlin/batchtest.kt
NicEastvillage: It does

falkecarlsen: Yes, JUnit5 does so in the testing-framwork. https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#fail-java.lang.String-java.lang.Throwable-

jenrik (approved):

PR #146, opened 2019-05-07
Codegen kotlin

jenrik:
This is a pile of changes and additions I have been working on the last week. They should probably have been split into multiple pull request, but so many of the changes are interdependent which makes it a hassle to do so, instead I suggest reviewers go through the commits one by one.


compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+class World(private val dims: List<Int>) {
+    var world = generateWorld(dims)
+
+    private fun generateWorld(dims: List<Int>): MutableList<Any> {
+        var x = dims[0]
+        if (dims.size == 1) {
+            return generateSequence {
```
NicEastvillage, line 144: Replace with
```
return Array(x) { 0 }.toMutableList()
```
This creates an array of size `x` that contains only 0's and then transforms it into mutable list

compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,195 @@
...
+                if (x >= 0) {
+                    0
+                } else {
+                    null
+                }
+            }.toMutableList()
+        } else {
+            val subDims = dims.subList(1, dims.size)
+
+            return generateSequence {
```
NicEastvillage, line 155: Replace with
```
return Array(x) { generateWorld(subDims) }.toMutableList()
```
At this point I have to say. I don't like the data structure `MutableList<Any>`. Lets make our own!

compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,195 @@
...
+                (pos[index] + p) >= dims[index] || (pos[index] + p) < 0
+            } else {
+                false
+            }
+        }.contains(true)
+        return if (overEdge) {
+            edge
+        } else {
+            world.getCell(*relPos.mapIndexed { index, p ->
+                val x = (pos[index] + p).rem(dims[index])
```
NicEastvillage, line 57: This should be abstracted. I suggest we make the function `wrap` defined as:
```kotlin
infix fun Int.wrap(divisor: Int) = (this % divisor).let { if (it < 0) it + divisor else it }
```

compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,195 @@
...
+                        )
+                    )
+                    x--
+                }
+            } else {
+                var x = subDims[0] - 1
+                while (x >= 0) {
+                    @Suppress("UNCHECKED_CAST")
+                    iterate(subDims.subList(1, subDims.size), pos + listOf(x))
+                    x--
```
NicEastvillage, line 123: `@Suppress("UNCHECKED_CAST")` is unnecessary.
`pos + listOf(x)` can be `pos + x` instead
... and if we want to be real fancy `pos + x--` and remove the following `x--`

compiler/runtime/src/main/kotlin/runtime/graphics.kt
```diff
@@ -0,0 +1,60 @@
...
+        val g = panel.graphics as Graphics2D
+
+        drawWorld(g)
+
+        Timer().scheduleAtFixedRate(object: TimerTask() {
+            override fun run() {
+                update()
+                drawWorld(g)
+            }
+        }, 0, tickrate)
```
NicEastvillage, line 38:
```suggestion
        }, 0, (1000f / tickrate).toLong())
```

NicEastvillage:
Remove all decrecated things. No one else relies on them anyway.

compiler/src/main/kotlin/visitors/codegen.kt
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    override fun visit(node: IfStmt): String {
+        val builder = StringBuilder()
+
+        node.conditionals.forEachIndexed { i, it ->
+            if(i > 0) {
+                builder.append(' ')
+            }
+            builder.append(visit(it))
```
NicEastvillage, line 551: Haven't tested, but isn't elif's missing the "else" part?

compiler/src/main/kotlin/visitors/codegen.kt
```diff
@@ -0,0 +1,650 @@
...
+        builder.append("""
+            }
+        """.trimIndent())
+
+        closeScope()
+
+        builder.append("\n\n")
+        builder.appendln(emitMain(
+            node.world,
+            node.body.filter { it is StateDecl }.map { it as StateDecl }.toList()
```
NicEastvillage, line 207:
```suggestion
            node.body.filterIsInstance<StateDecl>().toList()
```

compiler/src/main/kotlin/visitors/codegen.kt
```diff
@@ -0,0 +1,650 @@
...
+        addMapping("randf", nextLabel())
+        addMapping("absi", nextLabel())
+        addMapping("absf", nextLabel())
+        addMapping("floor", nextLabel())
+        addMapping("ceil", nextLabel())
+        addMapping("root", nextLabel())
+        addMapping("pow", nextLabel())
+
+        var stateCounter = 0
+        node.body.filter { it is StateDecl }.map { it as StateDecl }.forEach {
```
NicEastvillage, line 167:
```suggestion
        node.body.filterIsInstance<StateDecl>().forEach {
```

jenrik:
Closed because the changes have been merged in other smaller pull requests

PR #147, opened 2019-05-08
Kotlin codegen and runtime

jenrik:
I have tried to extract just the code generation and runtime from #146


compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,136 @@
...
+                (pos[index] + p) >= dims[index] || (pos[index] + p) < 0
+            } else {
+                false
+            }
+        }.contains(true)
+        return if (overEdge) {
+            edge
+        } else {
+            world.getCell(*relPos.mapIndexed { index, p ->
+                val x = (pos[index] + p).rem(dims[index])
```
NicEastvillage: Wrap? :(

compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Implements a world view capable of having edge states and wrapping dimensions
+ */
+class MultiWorldType(private val edge: Int, private val dimTypes: List<DimensionType>): IWorldType {
+    override fun getWorldView(world: World, dims: List<Int>, pos: List<Int>): IWorldView {
+        return MultiWorldView(world, dims, pos, edge, dimTypes)
+    }
+}
+
+/**
+ * Run a cellular automator over every cell in a world.
```
NicEastvillage:
```suggestion
 * Runs a cellular automata by updating every cell in a world in discrete timesteps.
```

compiler/runtime/src/main/kotlin/runtime/driver.kt
```diff
@@ -0,0 +1,151 @@
...
+        worldCurrent = worldNext
+        worldNext = temp
+    }
+}
+
+class World(private val dims: List<Int>) {
+    /**
+     * World is implemented as an N dimensional MutableList.
+     * Because the N isn't know when the code is written, an exact type can't be specified.
+     * There for Any is used to allow for both MutableList and Int.
```
NicEastvillage:
```suggestion
     * Therefore Any is used to allow for both MutableList and Int.
```

compiler/runtime/src/main/kotlin/runtime/graphics.kt
```diff
@@ -0,0 +1,60 @@
...
+                drawWorld(g)
+            }
+        }, 0, (1000F / tickrate).toLong())
+    }
+
+    fun drawWorld(g: Graphics2D) {
+        var x = 0
+        while(x < worldConfig.dims[0]) {
+            var y = 0
+            while(y < worldConfig.dims[1]) {
```
NicEastvillage: For-loops?

NicEastvillage (changes requested):
2 spelling errors. One whille-loop that should be a for-loop

compiler/runtime/src/main/kotlin/runtime/driver.kt
jenrik: That isn't the point i tried to express, I tried to make it clear that we run the program over each cell which result in the next configuration

NicEastvillage (dismissed):

NicEastvillage (approved):

Issue #140, opened 2019-05-07
No typecheck for become


MathiasMehl:
No typechecking for become, so become "functionName"; is fine, become 5; is also fine.

state ident (0, 0, 0) {
    become 5/0;
}

PR #148, opened 2019-05-08
Become typecheck

### LINKED ISSUE #135: Edge default state never checked
As of right now, the following program compiles without error, even though the default state for the edge is non-existing. Add test files for this :)

```c
world {
    size = 100 [edge], 200 [wrap];
    edge = abc;
}

state ident (0, 0, 0) {
    become ident;
}
```

MathiasMehl:
- Error will be added if the statemnt `become` is followed by something else than a state (0613094)
- Warning added if default edge is declared, but no dimension is [EDGE] (e8c755d)
- Error added if default edge is not a state (1d0b924)

Resolves #135 #140


falkecarlsen (approved):
Nice

PR #149, opened 2019-05-09
Addition of EqualityComparisonExpr so equality can compare all types

### LINKED ISSUE #131: Equality `==` is a NumericComparison, but should be able to compare anything
For instance
`true == false`
`Alive == Dead`
`{Alive, Dead} == {x, y}`
is not possible right now.

Same issue for inequality.

NicEastvillage:
See the issue. Resolves #131

### DIFF
--- a/compiler/src/main/kotlin/visitors/typechecker.kt
+++ b/compiler/src/main/kotlin/visitors/typechecker.kt
@@ -327,6 +327,50 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
         node.setType(symbolTableSession.getSymbolType(node.spelling)!!)
     }
 
+    override fun visit(node: EqualityComparisonExpr) {
+        super.visit(node)
+
+        val lt = node.left.getType()
+        val rt = node.right.getType()
+
+        when {
+            // Could not determine type of a child node, so we can't raise any errors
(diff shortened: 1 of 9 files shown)


compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -327,6 +327,50 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+            }
+
+            // int-to-float conversion for right child
+            lt == FloatType && rt == IntegerType -> {
+                node.right = IntToFloatConversion(node.right)
+            }
+
+            // state-array-to-neighbourhood conversion for left child
+            rt == LocalNeighbourhoodType && lt is ArrayType && lt.subtype == StateType -> {
+                node.right = StateArrayToLocalNeighbourhoodConversion(node.right)
```
falkecarlsen: Whoops. Should be left

falkecarlsen (changes requested):

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: Whoops 😂 👍 💯

falkecarlsen (approved):
Nice, looks good. *Don't really enjoy the 'awesome' comments though*

PR #151, opened 2019-05-09
Added commandline argument parsing

jenrik:


compiler/src/main/kotlin/main.kt
```diff
@@ -10,67 +11,125 @@ import org.antlr.v4.runtime.CharStream
...
+class Arguments(parser: ArgParser) {
+    val verbose by parser.flagging(
+        "-v", "--verbose",
+        help = "enable verbose mode"
+    )
+
+    val logLevel by parser.storing(
+        "--log-level",
+        help = "level of details to log [silent,minimal,debug,verbose]"
+    ).default("silent").addValidator {
```
NicEastvillage: Let's define each level at some point. Default should not be "silent" as the typical user should know what compile errors and warnings the compiler found.

compiler/src/main/kotlin/main.kt
```diff
@@ -10,67 +11,125 @@ import org.antlr.v4.runtime.CharStream
...
+        ).parseInto(::Arguments)
+            .run {
+                val settings = CompilerSettings(
+                    target = when(target) {
+                        "kotlin" -> CompileTarget.KOTLIN
+                        "cellmata" -> CompileTarget.CELLMATA
+                        else -> throw SystemExitException("Internal error", 255)
+                    },
+                    source = Paths.get(source),
+                    output = Paths.get(outputDir)
```
NicEastvillage: What about log level and graph phases?

compiler/src/main/kotlin/main.kt
```diff
@@ -87,60 +146,83 @@ fun compile(source: CharStream, settings: CompilerSettings): CompilerData {
...
 /**
  * Production compile function, outputs useful information to user on errors in source and possibly the compiler itself
  */
-fun prodCompilation(sourcePath: Path, settings: CompilerSettings) {
+fun prodCompilation(settings: CompilerSettings) {
     try {
-        compile(CharStreams.fromPath(sourcePath), settings)
-        ErrorLogger.printAllWarnings(if (settings.veryVerbose) Files.lines(sourcePath) else null)
+        compile(CharStreams.fromPath(settings.source), settings)
+        ErrorLogger.printAllWarnings(if (settings.logLevel == LogLevel.DEBUG) Files.lines(settings.source) else null)
```
NicEastvillage: Ideally we want to print warnings if the log level is minimal *or greater*.

NicEastvillage (changes requested):

compiler/src/main/kotlin/main.kt
NicEastvillage: Log levels:
- Silent: Nothing except exit code
- Minimal: Errors and warnings without pointers
- Debug: Errors and warnings with pointers
- verbose: Errors and warnings with pointers plus whatever else we got

jenrik: I agree that we should probably move the default log level to minimal.

My suggestions for levels:

* Silent: Nothing except exit code
* Minimal: Error intended for the user, i.e. syntax and semantic error, and warnings
* Debug: Debugging information intended for compiler developers, i.e. phase startet, and other internal information.
* Verbose: **Everything**, entering and leaving visitor functions, caught exceptions, ...

compiler/src/main/kotlin/main.kt
jenrik: Thats a bug

NicEastvillage: Added these myself in e175f6d

compiler/src/main/kotlin/main.kt
NicEastvillage: Added this myself in e175f6d

compiler/src/main/kotlin/main.kt
NicEastvillage: Sure!

NicEastvillage (approved):

PR #153, opened 2019-05-09
Added info about multi-states. Adjusted printers accordingly.

### LINKED ISSUE #150: The size of a multi-state-decl is not saved in StateDecl
We currently have no information in the StateDecl class about multi-state-decl

NicEastvillage:
This adds information about how many states a multi-state-decl declares. The sanity checker makes sure that it is not 0 or negative. Pretty printer and grapher have been adjusted accordingly.

It is still not possible to lookup multi-state, e.g. `Alive[2]`.

Resolves #150


falkecarlsen (approved):
Looks good.

PR #154, opened 2019-05-09
Fixed shadowing and function names having a type.

### LINKED ISSUE #144: The type of a function's name is its return type, not function
Currently this will compile:
```
state Alive (0, 0, 0) {
    let x = f + f;
}

function f () int {
    return 0;
}
```
Because the type of "f" is "int". This is caused by this line: https://github.com/d409f19/cellmata/blob/master/compiler/src/main/kotlin/ast/symboltable.kt#L95

NicEastvillage:
See issue #144 for details. While fixing this I discovered that shadowing didn't work when it's type changed.
Resolves #144


compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -324,7 +325,22 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+            ErrorLogger += TypeError(node.ctx, "Found function name '${node.spelling}', but the function is not called.")
+        } else {
+            node.setType(
+                when (decl) {
+                    is StateDecl -> StateType
+                    is NeighbourhoodDecl -> LocalNeighbourhoodType
+                    is AssignStmt -> decl.expr.getType()
+                    is ConstDecl -> decl.expr.getType()
+                    is FunctionArgument -> decl.type
+                    else -> throw AssertionError()
```
falkecarlsen: Why no error message for this?

falkecarlsen (changes requested):
Looks good, just one question.

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: It should never happen. Guess I can give it an error message anyway, so if we change something, we get a nice warning.

NicEastvillage: It should never happen. Guess I can give it an error message anyway, so if we change something, we get a nice warning.

falkecarlsen (approved):

PR #155, opened 2019-05-10
Symboltest minor fixes

falkecarlsen:
Turns out that `symbolRedefinitionFailTest` was defective, it didn't assert for some errors, but only asserted for each error, which is trivially correct when no errors exist.


NicEastvillage (dismissed):
Nice

NicEastvillage (approved):

PR #156, opened 2019-05-10
Added builtin functions' code gen

NicEastvillage:
Added builtin functions' implementation!

Depends on PR #151


compiler/src/main/kotlin/visitors/codegen.kt
```diff
@@ -188,15 +188,15 @@ class KotlinCodegen : ASTVisitor<String> {
 
     override fun visit(node: RootNode): String {
         // ToDo write implementation of builtins
```
falkecarlsen: I think this is done

falkecarlsen (changes requested):
Looks nice, one todo-should be removed though.

compiler/src/main/kotlin/visitors/codegen.kt
falkecarlsen: Seems fixed in 3e50348

falkecarlsen (approved):

PR #158, opened 2019-05-13
LookupExpr can lookup both arrays, multi-states, and neighbourhoods

NicEastvillage:


compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -54,16 +54,31 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
 
-        val arrayType = node.arr.getType()
+        val type = node.target.getType()
 
-        if (arrayType is ArrayType) {
-            node.setType(arrayType.subtype)
-        } else {
-            ErrorLogger += TypeError(node.ctx, "Cannot lookup in expression of type $arrayType")
-            node.setType(UndeterminedType)
+        when (type) {
```
falkecarlsen, line 62: Maybe throw assignment above into when-body :woman_shrugging:

falkecarlsen (approved):
Looks good, could simplify a when-statement though.

compiler/src/main/antlr/Cellmata.g4
```diff
@@ -101,7 +101,7 @@ expr : '#' # stateIndexExpr
     | PAREN_START value=expr PAREN_END # parenExpr
     | value=array_value_sized # arraySizedValueExpr
     | value=array_value_literal # arrayLiteralExpr
-    | value=expr SQ_BRACKET_START index=expr SQ_BRACKET_END # arrayLookupExpr
+    | target=expr SQ_BRACKET_START index=expr SQ_BRACKET_END # lookupExpr
```
jenrik, line 104: I like the change from `value` to `target`, it's easier to remember it as the target of the lookup operation

compiler/src/main/kotlin/visitors/prettyprinter.kt
```diff
@@ -104,7 +104,7 @@ class PrettyPrinter : BaseASTVisitor() {
      */
     override fun visit(node: StateDecl) {
         // Print signature of state. TODO: must take multi-state-declaration into account when implemented
-        stringBuilder.appendln("state ${node.ident} (${node.red}, ${node.green}, ${node.blue}) {")
+        stringBuilder.appendln("state ${node.ident}[${node.multiStateCount}] (${node.red}, ${node.green}, ${node.blue}) {")
```
jenrik, line 107: Have we fully agreed that single-state is equivalent to a multi-state with size 1?

compiler/src/main/kotlin/visitors/typechecker.kt
```diff
@@ -54,16 +54,31 @@ class TypeChecker(symbolTable: Table) : ScopedASTVisitor(symbolTable = symbolTab
...
+            is ArrayType -> {
+                node.lookupType = LookupExprType.ARRAY
+                node.setType(type.subtype)
+            }
+
+            LocalNeighbourhoodType -> {
```
jenrik, line 68: Should this case and `StateType` be prefixed with `is`?

jenrik (changes requested):

compiler/src/main/kotlin/visitors/prettyprinter.kt
NicEastvillage: We haven't talked about it, but functionally they must be equivalent.

compiler/src/main/kotlin/visitors/typechecker.kt
NicEastvillage: No. `is ArrayType` is neccessary because ArrayType is a class, `LocalNeighbourhoodType` and `StateType` are objects, so we can just test if the type is one of those specific objects.

jenrik (approved):

PR #163, opened 2019-05-13
Progressive scan and inline grid-swap 

falkecarlsen:
- The interpreter now iterates the grid in reverse, which flips the rendering-_line_ along the diagonal, such that each update happens in horisontal lines and descending from top to bottom. All driver-related references seem good and I don't believe this will break any other dependencies.

- Inline swapping of grid and next-grid by using Kotlin's `also`-operator, seems to improve performance a bit on my machine, but without metrics, it's hard to say definitively whether it hurts or helps.

- Also reformatted entire `interpreter.kt`, only whitespace is fixed.

### DIFF
--- a/compiler/src/main/kotlin/interpreter/interpreter.kt
+++ b/compiler/src/main/kotlin/interpreter/interpreter.kt
@@ -109,7 +109,8 @@ class Interpreter(val rootNode: RootNode) : ASTVisitor<Any> {
         Timer().scheduleAtFixedRate(object : TimerTask() {
             override fun run() {
 
-                // Iterate the grid from top to bottom, imitating a progressive scan
+                // Iterate the grid from top to bottom, imitating a progressive scan,
+                // find each cell's new state, and draw the new state
                 for ((y, column) in grid.withIndex()) {
                     for ((x, state) in column.withIndex()) {
                         // Find new state by executing the state's logic. If the logic does not return a StateDecl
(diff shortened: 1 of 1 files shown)


NicEastvillage:
~~I am sorry to break it to you, but you haven't "*Reworked run-driver to imitate progressive scan*". Try inserting a `Thread.sleep(1)` in the driver. You simply renamed the variables.~~

And doing inline swap saves us about 0 JVM instructions. It is just addresses temporarily hold in a register. I can check the exact difference if you want. But it is fancy!

falkecarlsen:
As you mentioned in chat, progressive scan does actually work :sunglasses:

NicEastvillage:
Yes. But only visually. The y-axis is now to the right. ~~I suggest you rerename things back to what they were, and only change the rendering, i.e. swap x and y in `g.fillRect(x * cellSize, y * cellSize, cellSize, cellSize)`.~~ To make the axis correct, you have to swap which dimension is added to x and y in `declareNeighbourhoods()`. I think this is the only place where we lookup in the grid and where the axis matter.

But anyway, this PR is only relevant for slow computers and doesn't *add* anything

falkecarlsen:
Progressive scan is definitely only a nice-to-have and its only function is to visually render the grid horisontally.  

I found [JMH](http://openjdk.java.net/projects/code-tools/jmh/) for benchmarking JVM-targeted programs down to nano-resolution but it's hardly worth the hassle.

falkecarlsen:
Closed since it's less important.

PR #164, opened 2019-05-14
Disallow function calls and neighbourhoods in constant declarations

NicEastvillage:
Neighbourhoods (and functions since they can mention neighbourhoods) are not allowed in constants, as no cell is being evaluated.

Resolves #164


jenrik:
As a long term fix for doing implementations of limited expressions, I suggest that we implement a stack of visited nodes such that you can see if you're working on a sub-node of a particular node

NicEastvillage:
That seems like a lot of effort to check if a node has been visited. If you were looking for a specific pattern in parent nodes, then it could be useful.

jenrik (approved):

PR #165, opened 2019-05-15
Added programs created in usability tests

MathiasMehl:
Added the programs that the users made during the usability tests


falkecarlsen (approved):
Cool

PR #167, opened 2019-05-16
Enable and disable relevant flow-warning test-programs

falkecarlsen:
As title says, these were previously not enabled nor disabled. Turns out the following `if-else-for.cell` program does not work:
```
world {
    size = 100 [wrap], 200 [wrap];
    tickrate = 1;
    cellsize = 5;
}

state ident (0, 0, 0) {
    become ident;
}

function f(int a) int {
    if (a == 1) {
        return a;
    } else {
        return a;
    }

    for (let x = a; x < a; x = x+1) {
        let y = 2;
    } 
    return a;
}
```


NicEastvillage (approved):
Weird that one doesn't work. Oh well.

falkecarlsen:
Made issue #168 too

PR #176, opened 2019-05-19
Added visit function for error nodes

jenrik:
Added visit function for ErroAST, ErrorDecl, ErrorStmt, and ErrorExpr.
All four default to throwing a UnreachableVisitorError.
Added an Error for when a visit function that should be unreachable is
hit.


NicEastvillage (approved):
Nice

PR #177, opened 2019-05-19
Generalized scope handling in visitors

jenrik:
Generalized scoping handling into BaseScopedASTVisitor in order to make
ScopedASTVisitor and SymbolChecker consistent with each other.

fixes #175

PR #178, opened 2019-05-19
Phase for making identifiers globally unique

jenrik:
Uniqueify replaces identifiers for all named nodes with a unique labels. This can in some cases remove the need to keep track scopes.

Depends on #177

### DIFF
--- a/compiler/src/main/kotlin/ast/ast.kt
+++ b/compiler/src/main/kotlin/ast/ast.kt
@@ -364,7 +364,7 @@ class NeighbourhoodDecl(
  */
 class FunctionArgument(
     ctx: SourceContext,
-    val ident: String,
+    var ident: String,
     val type: Type = UncheckedType
 ) : AST(ctx)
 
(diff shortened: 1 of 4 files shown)


NicEastvillage:
Can you give a small explaination of why this is needed?

You haven't added the phase to the compilation function(s). Is that intended?

And CI fails.

jenrik:
I need this as a prerequisite for the webassembly target. WebAssembly has very little in terms of scoping, for instance it has scopes for functions, but not for any of the control structures. This forces me to move variables of for loops up into the function scope, and having unique names would solve the annoying case where a variable in for loops initialization part shadows another variable in the function/state.

jenrik:
I should also note that I have run this phase through the tests, and it passes all except the few tests that check for specific variable names which it fails for obvious reasons.

NicEastvillage:
Sounds reasonable, but I wont approve before CircleCI passes. And I also don't want to submit this for our exam, so I wont approve it before that has been done.

PR #180, opened 2019-05-23
Adds life-like CA

MathiasMehl:
- Life-like CA are basically other versions of CGOL, just with other birth/survival rules.

- Adds the notable life-like rules as per https://en.wikipedia.org/wiki/Life-like_cellular_automaton

- Moves CGOL into this folder aswell. 

- Adds my edit of @NicEastvillage rock-paper-scissors to this folder aswell


falkecarlsen (approved):
Cool

